PR #3, opened 2023-04-17
Start on rearchitecting

njooma:
Includes camera stream, sensor, movement sensor, base, and arm. 

Is still usable using `Viam.instance`.


README.md
```diff
@@ -4,7 +4,7 @@ Build and connect to robots with Flutter
 ## (In)stability Notice
 
 > **Warning**
-> This is an alpha release of the Viam Flutter SDK. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
+> This is an alpha release of the Viam Flutter SDK. The only guarantee is that there **WILL** be breaking changes to the API. We will enumerate these changes in the release notes.
```
clintpurser, line 7: Love this guarantee haha 👍

lib/src/components/arm/client.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  @override
+  Future<bool> isMoving({Map<String, dynamic>? extra}) async {
+    final response = await _client.isMoving(IsMovingRequest(name: name));
+    return response.isMoving;
+  }
+
+  @override
+  Future<void> moveToJointPositions(JointPositions positions, {Map<String, dynamic>? extra}) async {
+    await _client.moveToJointPositions(MoveToJointPositionsRequest(name: name, positions: positions));
```
clintpurser, line 28: just wondering, Will we want a way to communicate that the `moveToJointPositions` was successful?

lib/src/components/arm/client.dart
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @override
+  Future<JointPositions> getJointPositions({Map<String, dynamic>? extra}) async {
+    final response = await _client.getJointPositions(GetJointPositionsRequest(name: name, extra: extra?.toStruct()));
+    return response.positions;
+  }
+
+  @override
+  Future<void> stop({Map<String, dynamic>? extra}) async {
```
clintpurser, line 43:
```suggestion
  Future<void> stop({Map<String, dynamic> extra}) async {
```

Is there a reason you marked these as being able to be null?

lib/src/components/arm/arm.dart
```diff
@@ -0,0 +1,39 @@
...
+  Future<bool> isMoving();
+
+  static ResourceName getResourceName(String name) {
+    return Arm.subtype.getResourceName(name);
+  }
+
+  static Arm fromRobot(RobotClient robot, String name) {
+    return robot.getResource(Arm.getResourceName(name));
+  }
+}
```
clintpurser, line 39: remind me again why we have the abstract arm class and then the armClient class?

lib/src/components/movement_sensor/movement_sensor.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  @override
+  Future<Map<String, dynamic>> getReadings({Map<String, dynamic>? extra}) async {
+    Map<String, dynamic> readings = {};
+    try {
+      Position pos = await getPosition(extra: extra);
+      readings["position"] = pos.coordinates;
+      readings["altitude"] = pos.altitude;
+    } catch (exception) {
+      // TODO: Check if the exception is of a specific type and ignore or rethrow
```
clintpurser, line 54: [minor nit] probably easy enough to print the exception here. maybe another todo or a quick followup.

```suggestion
      // TODO: Check if the exception is of a specific type and ignore or rethrow
       print('Error: $exception');
```

lib/src/domain/resource/model/viam_resource_name.dart
```diff
@@ -24,6 +24,9 @@ class ViamResourceName {
         subtype,
         name,
       );
+
+  @override
+  String toString() => '${namespace}:${type}:${subtype}/${name}';
```
clintpurser, line 29: Nice use of the arrow and string interpolation.

clintpurser (approved):
LGTM, I asked a few questions for my own understanding or minor suggestions, but nothing is a blocker.

lib/src/components/arm/client.dart
njooma: I made these optional because they are optional in the proto definitions, so having this be nullable makes it super easy to call these functions from clients/servers: `arm.stop(request.extra)` instead of `arm.stop(request.extra ?? {})`

PR #4, opened 2023-04-24
RSDK-2827 - board client wrapper & example

clintpurser:
Adds initial cut at client wrapper for board.

Also adding example board view to sample app.

Screenshot 2023-04-24 at 4 41 44 PM Screenshot 2023-04-24 at 4 41 46 PM

https://user-images.githubusercontent.com/22106496/234115723-d469984f-a5d9-4924-889c-615e38ac5478.MOV


example/viam_example_app/ios/Podfile.lock
```diff
@@ -51,4 +51,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 775997f741c536251164e3eacf6e34abf2eb7a17
 
-COCOAPODS: 1.12.0
+COCOAPODS: 1.12.1
```
clintpurser, line 54: `Drive by`: this was blocking me from building I think.

example/viam_example_app/pubspec.yaml
```diff
@@ -32,6 +32,7 @@ dependencies:
     sdk: flutter
   viam_sdk:
     path: ../../
+  flutter_webrtc: ^0.9.26
```
clintpurser, line 35: `Drive by`: this was stopping me from building, i think. Although it may have been with how I was trying to build.

lib/src/components/board/client.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:grpc/grpc_connection_interface.dart';
+
+import '../../gen/component/board/v1/board.pbgrpc.dart';
+import '../../utils.dart';
+import 'board.dart';
+
+class BoardClient extends Board {
+  ClientChannelBase _channel;
```
clintpurser, line 8:
```suggestion
  ClientChannelBoard _channel;
```

clintpurser:
I just noticed some things I missed. Marking as draft...

example/viam_example_app/pubspec.yaml
njooma: this shouldn't be needed -- flutter_webrtc is included with he viam-sdk package

clintpurser: i'll take another look.

clintpurser: yup you're right, that was my mistake.

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,25 @@
...
+
+  Future<void> setGPIO(String pin, bool high, {Map<String, dynamic>? extra});
+
+  // TODO: add rest of board methods
+
+  static ResourceName getResourceName(String name) {
+    return Board.subtype.getResourceName(name);
+  }
+
+  static Board fromRobot(RobotClient robot, String name) {
```
benjirewis, line 22: What's the point of this method? Can users just call [`getResource`](https://github.com/viamrobotics/viam-flutter-sdk/blob/main/lib/src/robot/client.dart#L73-L75) directly and provide the type? I'm worried we'll forget to implement this in every resource. If we want it for convenience, is there a way to assert at compile time that every extension of `Resource` implemented `getResourceName` and `fromRobot`?

benjirewis (commented):
know next-to-nothing about dart/flutter, but just a question

lib/src/components/board/board.dart
clintpurser: Great point. There are probably multiple ways to handle that enforcement, the first that comes to mind is moving the definition of this method to `Resource` and have each class that inherits override it. 

Are you ok if I handle addressing this concern and changing this pattern in a follow up PR?

benjirewis: > the first that comes to mind is moving the definition of this method to Resource and have each class that inherits override it.

that SGTM. And yeah feel free to address in follow-up PR.

benjirewis (dismissed):

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
...
+class BoardStatus {
+  final Map<String, int> analogs;
+  final Map<String, int> digitalInterrupts;
+
+  BoardStatus(this.analogs, this.digitalInterrupts);
+
+  factory BoardStatus.fromProto(common.BoardStatus pbBoardStatus) {
+    BoardStatus boardStatus = BoardStatus(Map<String, int>(), Map<String, int>());
+    pbBoardStatus.analogs.forEach((key, value) => boardStatus.analogs[key] = value.value);
+    pbBoardStatus.digitalInterrupts.forEach((key, value) => boardStatus.digitalInterrupts[key] = value.value.toInt());
```
clintpurser, line 15: in the TS SDK these values are mapped in the [client](https://github.com/viamrobotics/viam-typescript-sdk/blob/main/src/components/board/client.ts#L63-L71), I thought it was a little cleaner to do it here in the BoardStatus factory constructor. open to suggestions though.

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
...
+    return boardStatus;
+  }
+}
+
+enum ViamPowerMode {
```
clintpurser, line 20: Naming this just `PowerMode` caused some problems and i'd need to add a prefix which would be used in a lot of places in this file if I wanted to keep the naming consistent. I chose to call this `ViamPowerMode` instead, because the prefix felt heavy handed. Open to going the other route if this isn't a naming scheme we want to follow.

lib/src/components/board/board.dart
clintpurser, line 15: `boardStatus.digitalInterrupts[key] = value.value as int);`

was giving me problems here, throwing errors around casting for some reason. while digging around i discovered that `Int64` has a [toInt() method](https://pub.dev/documentation/fixnum/latest/fixnum/Int64/toInt.html) - this may be more preferable to use rather than casting ourselves. Something to think about, I will put up a follow up converting the code base to use this and that discussion can be continued there if need be.

lib/src/components/board/board.dart
benjirewis: What's the "naming problem"? Is there another variable called PowerMode?

benjirewis (dismissed):

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
...
+  Future<BoardStatus> status({Map<String, dynamic>? extra});
+
+  /// Set the high/low state of the given pin of a board.
+  Future<void> setGPIO(String pin, bool high, {Map<String, dynamic>? extra});
+
+  /// Get the high/low state of the given pin of a board.
+  Future<bool> getGPIO(String pin, {Map<String, dynamic>? extra});
+
+  /// Get the duty cycle of the given pin of a board.
+  Future<double> pWM(String pin, {Map<String, dynamic>? extra});
```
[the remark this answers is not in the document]
clintpurser: this is a good callout, I agree this is not very consistent with the use of `get` - I was following the naming in [board.pbgrpc.dart](https://github.com/viamrobotics/viam-flutter-sdk/blob/main/lib/src/gen/component/board/v1/board.pbgrpc.dart#L87-L143) It made sense to match what was provided by the protobuf spec, and I'm not sure how much the code ought to deviate from that.

lib/src/components/board/board.dart
clintpurser: I agree, it was confusing for me. I'm having some more ideas around this I want to dig into to understand 👀 - wondering if the `as int` was trying to caste more of the statement than what I wanted, but that is just a hunch.

lib/src/components/board/board.dart
clintpurser: this is the error btw https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=ambiguous_import#ambiguous_import

```
The name 'PowerMode' is defined in the libraries 'package:viam_sdk/src/components/board/board.dart' and 'package:viam_sdk/src/gen/component/board/v1/board.pbenum.dart'.
Try using 'as prefix' for one of the import directives, or hiding the name from all but one of the imports.
```

clintpurser: I looked into adding the prefixes again, thinking i may be able to only add it to one, but it still insists I add a prefix to both, which would mean adding a prefix to everything coming into client.dart from board.dart, which is pretty much everything. I think I am back to where I was before of feeling that is heavy handed and may want to punt this decision a bit to think about where to move the enum to avoid this or some other solution 🤔

lib/src/components/board/board.dart
clintpurser: ☝️ that is what it was. refer to the current version, but wrapping the Int64 and `as int` in `()` made it work as expected. I wasn't wrapping it before and it was trying to convert the whole statement to int.

lib/src/components/board/board.dart
benjirewis: Sounds good to me. That's some helpful error messaging from `dart`.

lib/src/components/board/board.dart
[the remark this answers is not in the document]
benjirewis: I think it's reasonable enough to use the names the dart protobuf compiler chose.

benjirewis (dismissed):

lib/src/components/board/board.dart
njooma: Why can't we just use the generated `PowerMode`? Why do we need to create our own?

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
...
+import '../../gen/common/v1/common.pb.dart' as common;
+import '../../robot/client.dart';
+
+class BoardStatus {
+  final Map<String, int> analogs;
+  final Map<String, int> digitalInterrupts;
+
+  BoardStatus(this.analogs, this.digitalInterrupts);
+
+  factory BoardStatus.fromProto(common.BoardStatus pbBoardStatus) {
```
njooma, line 12: what does `factory` do here? Without the `factory`, is it not just a named initializer?

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
+import 'package:viam_sdk/src/resource/base.dart';
+
+import '../../gen/common/v1/common.pb.dart' as common;
+import '../../robot/client.dart';
+
+class BoardStatus {
+  final Map<String, int> analogs;
+  final Map<String, int> digitalInterrupts;
+
+  BoardStatus(this.analogs, this.digitalInterrupts);
```
njooma, line 10: Should we make this a `cost` initializer since all attributes are final?

lib/src/components/board/board.dart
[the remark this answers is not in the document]
njooma: I personally think they should be `get` for consistency within the native language. IMO SDKs should have better ergonomics than the generated code, so we can make the necessary adjustments for consistency

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,78 @@
...
+  @override
+  String name;
+
+  Board(this.name);
+
+  /// Get the status of the Board.
+  Future<BoardStatus> status({Map<String, dynamic>? extra});
+
+  /// Set the high/low state of the given pin of a board.
+  Future<void> setGPIO(String pin, bool high, {Map<String, dynamic>? extra});
```
njooma, line 43: While I prefer having `GPIO` fully capitalized, I think it goes against the [flutter style guide](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words)

lib/src/components/board/board.dart
clintpurser: I was trying to remove any direct references to pb/generated code. I guess it being an enum opens the door to passing in the enum value or something. I could look into that.

lib/src/components/board/board.dart
clintpurser: Good callout, you're right. I think there might be a way to set up some static analysis for stuff like this too 👀

lib/src/components/board/board.dart
clintpurser: yes we should. Good catch, thanks!

lib/src/components/board/board.dart
njooma: I don't think we need to remove direct refs to generated code, as long as the generated code fits within the paradigm of the native language. So I think using the generated enum would be ok

lib/src/components/board/board.dart
njooma: Oh cool. This was more of a question for me to understand what `factory` means hahaha but glad to have helped make this more idiomatic

lib/src/components/board/board.dart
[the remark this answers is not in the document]
clintpurser: Thanks for the conversation around this guys, I've talked with a couple people offline about this as well.

I'm still dialing my understanding about how we want to design our code around the PB generations, so this is really helpful!

My new understanding is that we want to honor what the protobuf generators do, but will defer to language specific guidelines and then further defer to consistency inside an sdk within itself.

I previously was holding the generated protobuf code as the source of truth, but I see now that that is incorrect. I will modify the client code to follow more typical dart standards. I'll also put some quick follow ups for renaming the other client's code as well.

one thing that I want to call out here is the use of `get` - the effective dart [design guide](https://dart.dev/guides/language/effective-dart/design) suggests to [AVOID starting a method name with get.](https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get) - I will follow that approach and look into if we can introduce some static analysis to enforce.

lib/src/components/board/board.dart
clintpurser: upon further inspection, the use of factory does provide the benefit of being able to place logic inside the constructor and return a specific object. I'll prefer to keep it a factory so I can continue to map the pb model inside it. Let me know if you feel strongly otherwise.

lib/src/components/board/board.dart
```diff
@@ -0,0 +1,73 @@
...
+  Board(this.name);
+
+  /// Get the status of the Board.
+  Future<BoardStatus> status({Map<String, dynamic>? extra});
+
+  /// Set the high/low state of the given pin of a board.
+  Future<void> setGpioState(String pin, bool high, {Map<String, dynamic>? extra});
+
+  /// Get the high/low state of the given pin of a board.
+  Future<bool> isGpioHigh(String pin, {Map<String, dynamic>? extra});
```
clintpurser, line 41: still iffy on this name. trying to follow https://dart.dev/guides/language/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable

benjirewis (dismissed):
LGTM! Thanks

lib/src/components/board/board.dart
clintpurser: ok I've decided i don't like this. reverting back to `gpio`

njooma (approved):
I like these changes. My only concern is that we've strayed far from the RDK names, but honestly I think that's ok. These are more idiomatic and make sense to me.

benjirewis (approved):

PR #10, opened 2023-05-03
RSDK-2828 - add documentation to motor.dart

clintpurser:
Adds documentation comments pulled and slightly modified from thepython sdk [motor.py](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/components/motor/motor.py)

### DIFF
--- a/lib/src/components/motor/motor.dart
+++ b/lib/src/components/motor/motor.dart
@@ -39,69 +39,38 @@ abstract class Motor extends Resource {
 
   /// Sets the "percentage" of power the [Motor] should employ between -1 and 1.
   /// When [powerPct] is negative, the rotation will be in the backward direction.
-  ///
-  /// Args:
-  ///     [powerPct] (double): Power between -1 and 1
-  ///         (negative implies backwards).
   Future<void> setPower(double powerPct, {Map<String, dynamic>? extra});
 
   /// Spin the [Motor] the specified number of [revolutions] at specified [rpm].
   /// When [rpm] or [revolutions] is a negative value, the rotation will be in the backward direction.
   /// Note: if both [rpm] and [revolutions] are negative, the motor will spin in the forward direction.
-  ///
-  /// Args:
-  ///     [rpm] (double): Speed at which the motor should move in rotations per minute
-  ///         (negative implies backwards).
-  ///     [revolutions] (double): Number of revolutions the motor should run for
-  ///         (negative implies backwards).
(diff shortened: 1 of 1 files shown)


lib/src/components/motor/motor.dart
```diff
@@ -33,22 +33,71 @@ abstract class Motor extends Resource {
 
   Motor(this.name);
 
-  Future<void> setPower(double powerPct, {Map<String, dynamic>? extra});
```
clintpurser, line 36: driveby: I changed this to from `powerPct` to `power` to match the python sdk / documentation. seemed minor enough to include this rename here.

This is the only code change in the PR.

cheukt: would keep at powerPct

clintpurser: ✅ back to `powerPct`

lib/src/components/motor/motor.dart
```diff
@@ -33,22 +37,71 @@ abstract class Motor extends Resource {
...
   Future<void> goFor(double rpm, double revolutions, {Map<String, dynamic>? extra});
 
+  /// Spin the [Motor] to the specified position (provided in revolutions from home/zero),
+  /// at the specified speed, in revolutions per minute.
+  /// Regardless of the directionality of the [rpm] this function will move
+  /// the [Motor] towards the specified position.
+  ///
+  /// Args:
+  ///     [rpm] (double): Speed at which the motor should rotate (absolute value).
+  ///     [positionRevolutions] (double): Target position relative to home/zero, in revolutions.
```
njooma, line 66: It's actually not super important to have these enumerated like this, because the flutter documentation generator is much better than the python one.

So while in python it's necessary to list out all args and return types, it actually looks bad in Flutter to do so:
> image

You can see that it doesn't format the args nicely and it just looks weird. 

Unlike in python where it formats like this:
> image

So I would remove all the `Args` and `Returns` lines.

njooma (changes requested):

lib/src/components/motor/motor.dart
clintpurser: Thanks for calling this out. I will remove those.

clintpurser: ✅ removed the Args and Returns from the documentation.

njooma (approved):

stuqdog (approved):

cheukt (approved):

PR #13, opened 2023-05-03
[RSDK-2826] Camera component

njooma:
Adds the camera component, along with mime types and custom Viam-RGBA decoding


example/viam_example_app/lib/main.dart
```diff
@@ -102,22 +102,34 @@ class _MyHomePageState extends State<MyHomePage> {
...
-
-  bool _isSensor(ResourceName rn) {
-    return rn.subtype == Sensor.subtype.resourceSubtype || rn.subtype == MovementSensor.subtype.resourceSubtype;
+  StreamClient _getStream(ResourceName name) {
+    return _robot.getStream(name.name);
   }
 
-  bool _isServo(ResourceName rn) {
-    return rn.subtype == Servo.subtype.resourceSubtype;
+  bool _isImplemented(ResourceName rname) {
```
clintpurser, line 109: Thanks for doing this, I think this is a great way to approach it.

[nit] 
```suggestion
  bool _hasScreen(ResourceName rname) {
```

or maybe
```suggestion
  bool _isNavigable(ResourceName rname) {
```

I get what you mean with `_isImplemented` but I wonder if that makes more sense to us internally vs a consumer of the SDK. Just a [nit] take it or leave it :)

example/viam_example_app/lib/main.dart
```diff
@@ -102,22 +102,34 @@ class _MyHomePageState extends State<MyHomePage> {
...
+      Camera.subtype.resourceSubtype,
+      MovementSensor.subtype.resourceSubtype,
+      Sensor.subtype.resourceSubtype,
+      Servo.subtype.resourceSubtype,
+    ].contains(rname.subtype);
   }
 
-  StreamClient _getStream(ResourceName name) {
-    return _robot.getStream(name.name);
+  Widget? _getScreen(ResourceName rname) {
```
clintpurser, line 118: its a simple example app, so maybe too big of a undertaking than what its worth, but perhaps making some named routes would be an elegant way to handle this and scale as we get a few more examples. Not too worried about it for now though.

example/viam_example_app/lib/main.dart
```diff
@@ -136,26 +148,9 @@ class _MyHomePageState extends State<MyHomePage> {
...
-                                  }
-                                  return SensorScreen(
-                                      sensor: resourceName.subtype == Sensor.subtype.resourceSubtype
-                                          ? Sensor.fromRobot(_robot, resourceName.name)
-                                          : MovementSensor.fromRobot(_robot, resourceName.name),
-                                      resourceName: resourceName);
-                                }))
+                    trailing: _isImplemented(resourceName) ? Icon(context.platformIcons.rightChevron) : null,
+                    onTap: () => _isImplemented(resourceName)
+                        ? Navigator.push(context, platformPageRoute(context: context, builder: (context) => _getScreen(resourceName)!))
```
clintpurser, line 153: 😗👌 ✨  great cleanup here!

example/viam_example_app/lib/screens/stream.dart
```diff
@@ -18,14 +22,59 @@ class StreamScreen extends StatefulWidget {
...
+      final convertFut = convertImageToFlutterUi(value.image ?? img.Image.empty());
+      convertFut.then((value) {
+        final pngFut = value.toByteData(format: ui.ImageByteFormat.png);
+        pngFut.then((value) => setState(() {
+              imageBytes = value;
+              _imgLoaded = true;
+            }));
+      });
+    });
+  }
```
clintpurser, line 49: [q] is there a reason you favored `then` over `async / await` in this method?

example/viam_example_app/lib/screens/stream.dart
```diff
@@ -18,14 +22,59 @@ class StreamScreen extends StatefulWidget {
...
+    ui.Image uiImage = fi.image;
+
+    return uiImage;
+  }
+
   @override
   initState() {
     super.initState();
-    _startStream();
+    // _startStream();
```
clintpurser, line 77: intended?

lib/src/media/image.dart
```diff
@@ -0,0 +1,188 @@
...
+    }
+    if (decoder == null) {
+      return null;
+    }
+    return decoder.decode(Uint8List.fromList(bytes));
+  }
+}
+
+class ViamImage {
+  /// THe mimetype of the image
```
clintpurser, line 69:
```suggestion
  /// The mimetype of the image
```

lib/src/media/image.dart
```diff
@@ -0,0 +1,188 @@
...
+  final MimeType mimeType;
+
+  /// The raw bytes of the image
+  final List<int> raw;
+
+  bool _imageDecoded = false;
+  img.Image? _image;
+
+  /// The decoded image, if available. If the [MimeType] is not supported, this will be null.
```
clintpurser, line 78:
```suggestion
  /// Get the decoded image, if available. If the [MimeType] is not supported, this will be null.
```

?

example/viam_example_app/lib/screens/stream.dart
njooma: Mostly because I don't know how to use async/await when called from UI code... :)

example/viam_example_app/lib/screens/stream.dart
njooma: Yup, I want to start stream when the user presses the button now, not on load. I will remove this line.

example/viam_example_app/lib/main.dart
njooma: I want to have this app be as simple as possible, so that anyone building a fresh flutter app can very easily follow along. I"m not sure how much complication routing adds, so it might be worth looking into. But this isn't supposed to be a *good* app, it's just supposed to show how to use our APIs. The *good* app comes later

lib/src/media/image.dart
```diff
@@ -0,0 +1,188 @@
...
+  img.Image? get image {
+    if (_imageDecoded) {
+      return _image;
+    }
+    _image = this.mimeType.decode(this.raw);
+    _imageDecoded = true;
```
zaporter-work, line 84: Should you check if _image is null before setting _imageDecoded? There seems to be no error handling if a new / unsupported mime type is passed.

lib/src/media/image.dart
njooma: I actually disagree with this change on semantics.

Personally, I use verbs (e.g. `get`) for function calls both in naming and in documentation. Since this is a property, using an active verb like `get` is misleading IMO.

zaporter-work (approved):
Looks similar to c++ repo.

example/viam_example_app/lib/screens/stream.dart
clintpurser: cool cool, ok no worries. We can refactor this later, to align patterns, not a big deal for now.

There are a couple of ways to handle it, the primary way could be with a [FutureBuilder<T>](https://api.flutter.dev/flutter/widgets/FutureBuilder-class.html) - it does require some reworking of the UI though. there are also some unofficial packages/widgets that we could use as well.

clintpurser (dismissed):

lib/src/media/image.dart
njooma: Perhaps `_imageDecoded` is a misnomer -- it's just a boolean to see if we've attempted decoding, not whether it was successful. Because if it was unsuccessful, we don't want to try decoding again (nothing will have changed, it'll be a waste of time). And if it was successful, we'll have the image. 

I can't simply do a `if (_image !== null)` test because it's perfectly OK for the `_image` to be `null`. So I have to check to see if we've attempted decoding already, and skip decoding a second time if yes

PR #16, opened 2023-05-08
align example and sdk lint & analysis settings

clintpurser:
This PR pulls in the lint and analysis settings from the SDK into the nested example project.

I like to run the example project from VS code by opening that project in its own window, I was running into inconsistencies in dart formatting the same file from the example project vs code window and the SDK vs code window, this resolves those inconsistencies in the auto formatting on save. As well as aligns the lint errors / warning that will be surfaced.

### DIFF
--- a/example/viam_example_app/.vscode/settings.json
+++ b/example/viam_example_app/.vscode/settings.json
@@ -1,3 +1,13 @@
 {
-    "cmake.configureOnOpen": false
+    "dart.lineLength": 140,
+    "[dart]": {
+        "editor.rulers": [
+            140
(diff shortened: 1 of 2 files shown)


clintpurser:
for reference, here is the SDK [settings.json](https://github.com/viamrobotics/viam-flutter-sdk/blob/main/.vscode/settings.json) & [analysis_options.yaml](https://github.com/viamrobotics/viam-flutter-sdk/blob/main/analysis_options.yaml). I just copied the whole file's contents for each.

njooma (approved):

PR #22, opened 2023-05-12
RSDK-3110 - Servo service & Testing

clintpurser:
This PR adds a Servo Service and unit testing for servo's.


lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:grpc/grpc.dart';
+import 'package:viam_sdk/src/components/servo/servo.dart';
+import 'package:viam_sdk/src/gen/common/v1/common.pb.dart';
+import 'package:viam_sdk/src/gen/component/servo/v1/servo.pbgrpc.dart';
+import 'package:viam_sdk/src/resource/manager.dart';
+import 'package:viam_sdk/src/utils.dart';
+
+class ServoService extends ServoServiceBase {
+  final ResourceManager _manager;
```
benjirewis, line 9: Why `final`?

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  ServoService(this._manager);
+
+  Servo _servoFromManager(String name) {
```
benjirewis, line 13: You may eventually want some level of genericism to allow grabbing any generic resource from the manager; that way you don't have to write the same `try/catch` many times with a resource not found error.

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      return _manager.getResource(Servo.getResourceName(name));
+    } catch (e) {
+      // TODO make rescource not found GRPC error
+      print('Exception details:\n $e');
+      rethrow;
```
benjirewis, line 19: Every time I read a flutter PR I learn a new keyword. Turns out dart has a whopping 61 keywords. Go has 25 which is abnormally low, but I thought rust was on the high end with 51. 😮 

Edit: swift has 71 and C++ has 95 so I don't even know anymore 🤷🏻

test/unit_test/components/servo_test.dart
```diff
@@ -0,0 +1,229 @@
...
+
+  group('Servo Service Tests', () {
+    late ClientChannel channel;
+    late FakeServo servo;
+    late ServoService service;
+    late Server server;
+    late String name;
+    final int testPosition = 42;
+
+    setUp(() async {
```
benjirewis, line 100: do `setUp` and `tearDown` get called before and after each `test` respectively?

benjirewis (commented):
Nice! Do these tests run in CI?

test/unit_test/components/servo_test.dart
clintpurser: They do. setUpAll() / tearDownAll() exist also if i wanted to run just once. I debated that option, but opted for a fresh start each time to limit unintended effects of testing on the same service / FakeServo.

lib/src/components/servo/service.dart
clintpurser: Thats a great call out, i'll look into that while building out the next one. Thank you!

lib/src/components/servo/service.dart
clintpurser: I want it to not be able to be set again. It may be overkill. I dont _think_ it will stop things from being added to it, just stops it from being pointed to a different object

lib/src/components/servo/service.dart
clintpurser: Wow, I had no idea there were so  many. Wild. Is that including meta annotations? 👀

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @override
+  Future<DoCommandResponse> doCommand(ServiceCall call, DoCommandRequest request) async {
+    // TODO: implement doCommand
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<GetPositionResponse> getPosition(ServiceCall call, GetPositionRequest request) async {
```
clintpurser, line 37: `get...` !! 

Thinking out loud here...

We don't have the same layer of abstraction here as we do on the client side, so this name has to stay unless we add a layer 🤔 is that worth it? We'll run into this more later with other resource services as well.

lib/src/components/servo/service.dart
benjirewis: I think `final` makes sense then.

lib/src/components/servo/service.dart
benjirewis: > Is that including meta annotations?

I _think_ so, yeah. I did some very fast googling for all of those keyword counts, so unclear how reputable my sources were 😆 .

lib/src/components/servo/service.dart
benjirewis: > so this name has to stay unless we add a layer

Confused what you mean here. Can you not use the method name `position` because of a name spacing issue?

benjirewis (dismissed):
LGTM!

benjirewis (approved):

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:grpc/grpc.dart';
+import 'package:viam_sdk/src/components/servo/servo.dart';
+import 'package:viam_sdk/src/gen/common/v1/common.pb.dart';
+import 'package:viam_sdk/src/gen/component/servo/v1/servo.pbgrpc.dart';
+import 'package:viam_sdk/src/resource/manager.dart';
+import 'package:viam_sdk/src/utils.dart';
```
njooma, line 6: Let's use relative imports since we're still within the SDK

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
...
+class ServoService extends ServoServiceBase {
+  final ResourceManager _manager;
+
+  ServoService(this._manager);
+
+  Servo _servoFromManager(String name) {
+    try {
+      return _manager.getResource(Servo.getResourceName(name));
+    } catch (e) {
+      // TODO make rescource not found GRPC error
```
cheukt, line 17: can you go ahead and make the error?

lib/src/components/servo/service.dart
```diff
@@ -0,0 +1,56 @@
...
+  Future<MoveResponse> move(ServiceCall call, MoveRequest request) async {
+    Servo servo = _servoFromManager(request.name);
+    await servo.move(request.angleDeg, extra: request.extra.toMap());
+    return MoveResponse();
+  }
+
+  @override
+  Future<DoCommandResponse> doCommand(ServiceCall call, DoCommandRequest request) async {
+    // TODO: implement doCommand
+    throw UnimplementedError();
```
cheukt, line 33: servo should have a docommand - was it not generated?

test/unit_test/components/servo_test.dart
```diff
@@ -0,0 +1,228 @@
...
+      service = ServoService(manager);
+
+      channel = ClientChannel(
+        'localhost',
+        port: 50051,
+        options: ChannelOptions(
+          credentials: const ChannelCredentials.insecure(),
+          codecRegistry: CodecRegistry(codecs: const [GzipCodec(), IdentityCodec()]),
+        ),
+      );
```
cheukt, line 116: might be good to refactor this bit of test utility code out

cheukt (changes requested):
some comments

cheukt (changes requested):
some comments

njooma (changes requested):
Solid! Just a small comment

lib/src/components/servo/service.dart
clintpurser: SGTM

test/unit_test/components/servo_test.dart
clintpurser: Good call out, thank you! I added a group wrapping these two groups and abstracted it out to there.

lib/src/components/servo/service.dart
clintpurser: I was putting off implementing it because there would be some extra work required to get doCommand working. I've done that now and included the changes in this PR.

lib/src/utils.dart
```diff
@@ -59,16 +59,37 @@ extension BoolValueUtils on bool {
...
+        }
+        if (value is bool) {
+          v = value.toValue();
+        }
+        if (value is List<ValueType>) {
+          v = value.toValue();
+        }
+        if (value is Map<String, dynamic>) {
+          v = value.toValue();
+        }
```
clintpurser, line 92: Extensions get set at compile time, while the way we were typing dynamics is a runtime decision, so it was unable to find `toValue()` for any of the types and would throw an error. We need to get the type at runtime so we know which version of toValue to call.

I don't like this, but it works, it would crash if a type other than the ones checked for was inside the map, so that needs to be accounted for.

reference: 
https://stackoverflow.com/questions/61921608/dart-use-generic-extension-method-get-nosuchmethoderror-class-xxx-has-no-in

clintpurser: it is confusing, i could probably explain it better too.

Here is the error it shows without having this type checking:

Screenshot 2023-05-22 at 11 32 38 AM

happy to try and explain offline

benjirewis: I think I understand better after an offline explanation. Definitely don't love the list of `ifs` here that do a sort of implicit casting to avoid the runtime complaints, but it doesn't seem like there's an obvious alternative given the static nature of extensions. Could you leave a comment in this method for posterity as to why this function is set up the way it is? And, can we only catch `NoSuchMethodError`s instead of all exceptions?

example/viam_example_app/test/widget_test.dart
```diff
@@ -1,30 +0,0 @@
...
-
-    // Tap the '+' icon and trigger a frame.
-    await tester.tap(find.byIcon(Icons.add));
-    await tester.pump();
-
-    // Verify that our counter has incremented.
-    expect(find.text('0'), findsNothing);
-    expect(find.text('1'), findsOneWidget);
-  });
-}
```
clintpurser, line 30: this was an artifact of the flutter starter project, it was causing the test suite to crash instead of just failing. 👋

lib/src/utils.dart
```diff
@@ -33,45 +35,37 @@ abstract class ValueType {
...
+        } else if (value is Map<String, dynamic>) {
+          result[entry.key] = value.toValue();
+          // ignore: type_check_with_null, prefer_void_to_null
+        } else if (value is Null) {
+          result[entry.key] = Value(nullValue: value);
+        } else {
+          throw GrpcError.invalidArgument('Unsupported type');
+        }
+      } catch (error) {
+        // TODO log error
```
clintpurser, line 68: Naveed has a PR up right now that adds a logging framework, once thats in i'll update this, or add in a quick follow up.

lib/src/utils.dart
```diff
@@ -33,45 +35,37 @@ abstract class ValueType {
...
-  }
-}
-
-extension BoolValueUtils on bool {
-  Value toValue() {
-    return Value(boolValue: this);
-  }
-}
-
+// TODO fix To toValue() for List<T?>
```
clintpurser, line 38: The scope of this PR has slowly been creeping, I'd like to address making List.toValue functional in a follow up PR.

lib/src/utils.dart
clintpurser: Benji and I spoke offline, since removing the `toValue()` calls for each type and calling the constructor directly in the Map method, we believe a comment isn't needed.

The only reason we'll catch the error is to log it, and then rethrow. so we don't need to target a specific error type.

lib/src/utils.dart
clintpurser: That PR was merged, i've rebased and added logging here.

lib/src/utils.dart
```diff
@@ -43,45 +48,37 @@ abstract class ValueType {
...
 extension ListValueUtils<T extends ValueType> on List<T> {
   Value toValue() {
-    return Value(listValue: ListValue(values: this.map((e) => e.toValue())));
+    return Value(listValue: ListValue(values: map((e) => e.toValue())));
   }
 }
 
 extension MapStructUtils on Map<String, dynamic> {
-  Struct toStruct({bool skipErrors = false}) {
```
njooma, line 77: Any reason to not let user's ignore errors?

njooma (approved):
Just a comment, shouldn't block merge

lib/src/utils.dart
clintpurser: no particular reason just didn't understand why that would be an option, I meant to ask you why it was included so I'm glad you saw this. 

Should I add it back?

njooma: Nah it's probably fine without it. We can always add it back later

cheukt (approved):

PR #24, opened 2023-05-19
Refactor WebRTC

njooma:
Refactor WebRTC code and implement DialOptions

### DIFF
--- a/example/viam_example_app/lib/main.dart
+++ b/example/viam_example_app/lib/main.dart
@@ -89,7 +89,9 @@ class _MyHomePageState extends State<MyHomePage> {
       _loading = true;
     });
     final robotFut = RobotClient.atAddress(
-        '<URL>', 443, RobotClientOptions.withSecret('<SECRET>'));
+      '<URL>',
+      RobotClientOptions.withLocationSecret('<SECRET>'),
+    );
 
     robotFut.then((value) {
       _robot = value;
--- a/lib/src/media/image.dart
+++ b/lib/src/media/image.dart
@@ -118,8 +118,6 @@ class _ViamRGBADecoder extends img.Decoder {
 
   @override
   img.Image? decodeFrame(int frame) {
-    Uint8List imageData;
-
     final image = img.Image(
       width: info.width,
       height: info.height,
--- a/lib/src/media/stream/client.dart
+++ b/lib/src/media/stream/client.dart
@@ -11,7 +11,6 @@ class StreamClient {
   // ignore: close_sinks
   final StreamController<MediaStream> _streamController = StreamController<MediaStream>.broadcast();
   MediaStream? _stream;
-  // final Map<String, StreamController<MediaStream>> _streams = {};
   StreamSubscription? _errorHandler;
 
   StreamClient(this._channel) : _client = StreamServiceClient(_channel) {
@@ -44,14 +43,13 @@ class StreamClient {
     return _streamController.stream;
   }
 
-  @override
   Future<void> add(String name) async {
     await _client.addStream(AddStreamRequest(name: _getValidSDPTrackName(name)));
   }
 
-  @override
   Future<void> remove(String name) async {
     await _client.removeStream(RemoveStreamRequest(name: _getValidSDPTrackName(name)));
     _stream = null;
+    await _errorHandler?.cancel();
   }
 }
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -13,6 +13,10 @@ import '../viam_sdk.dart';
 class RobotClientOptions {
   DialOptions dialOptions;
 
+  RobotClientOptions() : dialOptions = DialOptions();
+
+  RobotClientOptions.withDialOptions(this.dialOptions);
+
   RobotClientOptions.withLocationSecret(String locationSecret)
       : dialOptions = DialOptions()..credentials = Credentials.locationSecret(locationSecret);
 }
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -23,7 +23,6 @@ class RobotClient {
   List<ResourceName> resourceNames = [];
   ResourceManager _manager = ResourceManager();
   final Map<String, StreamClient> _streams = {};
-  Viam? viam;
 
   RobotClient._();
 
@@ -37,7 +36,6 @@ class RobotClient {
 
   static Future<RobotClient> withViam(Viam viam) async {
     var client = RobotClient._();
-    client.viam = viam;
     client.channel = viam.channel;
     client._client = RobotServiceClient(client.channel);
     await client.refresh();
@@ -45,8 +43,8 @@ class RobotClient {
   }
 
   Future<void> refresh() async {
-    ResourceNamesResponse response = await this._client.resourceNames(ResourceNamesRequest());
-    if (response.resources == this.resourceNames) {
+    ResourceNamesResponse response = await _client.resourceNames(ResourceNamesRequest());
+    if (response.resources == resourceNames) {
       return;
     }
     final manager = ResourceManager();
@@ -63,14 +61,14 @@ class RobotClient {
         continue;
       }
     }
-    this.resourceNames = response.resources;
-    if (this._manager.resources != manager.resources) {
-      this._manager = manager;
+    resourceNames = response.resources;
+    if (_manager.resources != manager.resources) {
+      _manager = manager;
     }
   }
 
   T getResource<T>(ResourceName name) {
-    return this._manager.getResource<T>(name);
+    return _manager.getResource<T>(name);
   }
 
   StreamClient getStream(String name) {
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -3,8 +3,6 @@ import 'package:viam_sdk/src/gen/robot/v1/robot.pbgrpc.dart';
 import 'package:viam_sdk/src/resource/manager.dart';
 import 'package:viam_sdk/viam_sdk.dart';
 
-import '../media/stream/client.dart';
-
 class RobotClientOptions {
   bool disableWebRtc = false;
   String locationSecret;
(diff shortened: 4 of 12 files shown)


example/viam_example_app/lib/main.dart
```diff
@@ -89,7 +89,9 @@ class _MyHomePageState extends State<MyHomePage> {
       _loading = true;
     });
     final robotFut = RobotClient.atAddress(
-        '<URL>', 443, RobotClientOptions.withSecret('<SECRET>'));
+      '<URL>',
+      RobotClientOptions.withLocationSecret('<SECRET>'),
```
clintpurser, line 93: ✨

lib/src/media/image.dart
```diff
@@ -118,8 +118,6 @@ class _ViamRGBADecoder extends img.Decoder {
 
   @override
   img.Image? decodeFrame(int frame) {
-    Uint8List imageData;
```
clintpurser, line 121: [q] why remove this here? is it a _driveby_ that you noticed was unused?

lib/src/media/stream/client.dart
```diff
@@ -44,14 +43,13 @@ class StreamClient {
     return _streamController.stream;
   }
 
-  @override
   Future<void> add(String name) async {
     await _client.addStream(AddStreamRequest(name: _getValidSDPTrackName(name)));
   }
 
-  @override
```
clintpurser, line 52: good catch on these 👍

lib/src/robot/client.dart
```diff
@@ -1,36 +1,35 @@
...
 
+import '../domain/web_rtc/web_rtc_client/web_rtc_client.dart';
+import '../gen/common/v1/common.pb.dart';
+import '../gen/robot/v1/robot.pbgrpc.dart';
 import '../media/stream/client.dart';
+import '../resource/base.dart';
+import '../resource/manager.dart';
+import '../resource/registry.dart';
+import '../rpc/dial.dart';
+import '../viam_sdk.dart';
```
clintpurser, line 11: I think we can add a lint rule to prefer relative imports. 👀  but maybe we don't want that because we'll want to do package imports in our example code. something to think about!

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,352 @@
...
+      try {
+        final candidateProto = ICECandidate(
+          candidate: candidate.candidate,
+          sdpMid: candidate.sdpMid,
+          sdpmLineIndex: candidate.sdpMLineIndex,
+        );
+        await signalingClient.callUpdate(CallUpdateRequest(uuid: uuid, candidate: candidateProto));
+      } catch (error, st) {
+        //TODO: Add error handling
+        _logger.e('Update ICECandidate error', error, st);
```
clintpurser, line 172: Thanks for setting up logger! 🙏

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,352 @@
...
+      if (!isResponseStreamInitialized) {
+        _logger.e('Updating before initialized');
+        throw Exception('Updating before initialized');
+      }
+      if (response.uuid != uuid) {
+        _logger.e('UUID mismatch in update. Got ${response.uuid}, want $uuid');
+        throw Exception('UUID mismatch in update. Got ${response.uuid}, want $uuid');
+      }
+
+      // await Future.delayed(const Duration(seconds: 1));
```
clintpurser, line 256: [minor] remove?

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,352 @@
...
+  String? externalAuthToEntity;
+
+  /// Allows a pre-authenticated client to dial with
+  /// an authorization header. Direct dial will have the access token
+  /// appended to the "Authorization: Bearer" header. WebRTC dial will
+  /// append it to the signaling server communication
+  ///
+  /// If enabled, other auth options have no affect. Eg. [authEntity], [credentials],
+  /// [externalAuthAddress], [externalAuthToEntity]
+  String? accessToken;
```
clintpurser, line 39: was there a particular reason you made these Strings`?` nullable?

IMO It adds a lot of boiler plate below to check if they're null and then if they're empty.

could we make them non nullable and initialize them to '' to skip those null checks?

```suggestion
  String authEntity = '';

  /// Used to authenticate the request.
  Credentials? credentials;

  /// Options to control how WebRTC is utilized in a dial attempt
  DialWebRtcOptions? webRtcOptions;

  /// The address to perform external auth at.
  String externalAuthAddress  = '';

  /// The entity to authenticate for after externally authenticating.
  String externalAuthToEntity  = '';

  /// Allows a pre-authenticated client to dial with
  /// an authorization header. Direct dial will have the access token
  /// appended to the "Authorization: Bearer" header. WebRTC dial will
  /// append it to the signaling server communication
  ///
  /// If enabled, other auth options have no affect. Eg. [authEntity], [credentials],
  /// [externalAuthAddress], [externalAuthToEntity]
  String accessToken  = '';
```

lib/src/robot/client.dart
```diff
@@ -1,36 +1,35 @@
...
 
 class RobotClient {
   late ClientChannelBase channel;
   late RobotServiceClient _client;
   List<ResourceName> resourceNames = [];
   ResourceManager _manager = ResourceManager();
-  Map<String, StreamClient> _streams = {};
-  late Viam viam;
+  final Map<String, StreamClient> _streams = {};
+  Viam? viam;
```
clintpurser, line 26: [minor] I think this is unused, remove?

a viam is referenced on lines 38-41, but that name is overloaded and the new one will be used and not this one.

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,352 @@
...
+  Credentials? signalingCredentials;
+
+  /// Allows a pre-authenticated client to dial with an authorization header to the signaling server.
+  /// This skips the Authenticate() request to the signaling server or external auth but does not skip the
+  /// AuthenticateTo() request to retrieve the credentials at the external auth endpoint.
+  ///
+  /// If enabled, other auth options have no affect. Eg. [signalingAuthEntity], [signalingCredentials],
+  /// [signalingExternalAuthAddress], [signalingExternalAuthToEntity]
+  String? signalingAccessToken;
+}
```
clintpurser, line 84: [minor] same suggestion as above, possibly remove nullable

clintpurser (changes requested):
Overall looks great! Just a few nits and minor suggestions.

lib/src/media/image.dart
njooma: Yup, unused

lib/src/rpc/dial.dart
njooma: Decided against doing this (because I liked the nil coalescing later on for convenience), so I instead made an extension on `String?` that will check for both null and empty. Makes things the cleanest :)

lib/src/rpc/dial.dart
njooma: Same as above

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,351 @@
...
+      ..id = 1
+      ..negotiated = true
+      ..ordered = true,
+  );
+
+  String? uuid;
+  final didConnect = Completer();
+  final didSetRemoteDesc = Completer();
+
+  // If tickleICE is enabled, set onIceCandidate handler
```
benjirewis, line 154:
```suggestion
  // If trickleICE is enabled, set onIceCandidate handler
```

benjirewis (commented):
Cool! 🧑‍🔧

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,351 @@
...
+
+      try {
+        final candidateProto = ICECandidate(
+          candidate: candidate.candidate,
+          sdpMid: candidate.sdpMid,
+          sdpmLineIndex: candidate.sdpMLineIndex,
+        );
+        await signalingClient.callUpdate(CallUpdateRequest(uuid: uuid, candidate: candidateProto));
+      } catch (error, st) {
+        //TODO: Add error handling
```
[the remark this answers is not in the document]
njooma: That todo is incorrect. We're not supposed to rethrow errors on call updates.

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,351 @@
...
+      ..negotiated = true
+      ..ordered = true,
+  );
+
+  String? uuid;
+  final didConnect = Completer();
+  final didSetRemoteDesc = Completer();
+
+  // If tickleICE is enabled, set onIceCandidate handler
+  if (!(options.webRtcOptions?.disableTrickleIce ?? config.disableTrickle)) {
```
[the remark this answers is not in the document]
njooma: `config.disableTrickle` was received by the server as an optional webrtc configuration. It shouldn't override `webrtcoptions.disable...` but it can be used as a backup in case the value the user passed in is empty

clintpurser (approved):
LGTM

lib/src/robot/client.dart
```diff
@@ -1,53 +1,50 @@
...
   RobotClient._();
 
-  static Future<RobotClient> atAddress(String url, int port, RobotClientOptions options) async {
+  static Future<RobotClient> atAddress(String url, RobotClientOptions options) async {
     var client = RobotClient._();
-    client.viam = Viam.instance();
-    await client.viam
-        .connect(url: url, port: port, secure: !options.insecure, disableWebRtc: options.disableWebRtc, payload: options.locationSecret);
-    client.channel = client.viam.channel;
+    client.channel = await dial(url, options.dialOptions);
```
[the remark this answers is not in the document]
njooma: It didn't before, but it does now :)

benjirewis (approved):
Nice, connecting local addresses will be helpful if we implement modules in dart.

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,405 @@
...
+  });
+  await peerConnection.setLocalDescription(offer);
+
+  final sdpJsonString = _convertSDPtoJsonString(await peerConnection.getLocalDescription());
+  final encodedSdp = _encodeSDPJsonStringToBase64String(sdpJsonString);
+  try {
+    callStream = signalingClient
+        .call(CallRequest(sdp: encodedSdp, disableTrickle: options.webRtcOptions?.disableTrickleIce ?? config.disableTrickle));
+  } catch (error, st) {
+    //TODO: Add error handling
```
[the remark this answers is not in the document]
njooma: This is a different TODO that I didn't catch earlier. But only the `CallUpdate` shouldn't rethrow, since it's not really required for everything to function

lib/src/rpc/dial.dart
```diff
@@ -0,0 +1,405 @@
...
+  }
+  final iceServers = config.additionalIceServers
+      .map((e) => {
+            'urls': e.urls,
+            'credential': e.credential,
+            'username': e.username,
+          })
+      .toList()
+    ..add({
+      'urls': 'stun:global.stun.twilio.com:3478?transport=udp',
```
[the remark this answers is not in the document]
njooma: I copied this over from what our friends over at DOR had done. Just tested it, seems like we don't need it

PR #26, opened 2023-05-23
Add Motor Service

clintpurser:
This PR adds the server side for Motor.

I will add testing in a follow up PR.


lib/src/components/motor/service.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  @override
+  Future<DoCommandResponse> doCommand(ServiceCall call, DoCommandRequest request) async {
+    Motor motor = _motorFromManager(request.name);
+    var result = await motor.doCommand(request.command.toMap());
+    return DoCommandResponse(result: result.toStruct());
+  }
+
+  @override
+  Future<GetPositionResponse> getPosition(ServiceCall call, GetPositionRequest request) async {
```
benjirewis, line 30: I forgot what we settled on for dart, `get` or no `get`?

benjirewis (approved):

lib/src/components/motor/service.dart
clintpurser: we decided no `get` for the clients, but for service wrappers we decided to mimic the proto exactly because we don't have the same inheritance pattern that allowed us to change the names as in the clients.

clintpurser:
@benjirewis This is the first PR with the CI turned on, it is catching a lot of style and other warnings that were already pre existing prior to these changes, my plan is that I am going to override it for now and address the issues or turn it off in a quick follow up.

maximpertsov (approved):

PR #27, opened 2023-05-23
Fix CI pipeline, apply misc idiomatic changes.

clintpurser:
revert to default lineLength lint rule.

make tests not fail on warnings. This ought to be removed once our warnings have been addressed.

Apply changes from `dart fix`


.vscode/settings.json
```diff
@@ -1,10 +1,4 @@
 {
-    "dart.lineLength": 140,
-    "[dart]": {
-        "editor.rulers": [
-            140
-        ],
```
clintpurser, line 6: my reasoning behind removing this is, if we want to leverage `dart format` in the future for our CI it will not honor this setting, so we'll be failing tests because of line length formatting issues. 

IMO for stuff like this i'd rather leverage the already written solutions instead of customizing our codebase, it will be one less thing we need to worry about.

clintpurser: most of the changes in this PR are bumping lines to not break 140 characters, FYI.

.github/workflows/test.yaml
```diff
@@ -19,12 +19,10 @@ jobs:
             
       - run: flutter --version
       - run: flutter pub get
-
-      - name: Verify formatting
-        run: dart format --output=none .
```
clintpurser, line 24: removing `dart format` for now. I think we'll want to add this back in soon, but for now it is failing.

.github/workflows/test.yaml
```diff
@@ -19,12 +19,10 @@ jobs:
...
+      - run: flutter pub get --directory=example/viam_example_app
 
       - name: Analyze project source
-        run: dart analyze --fatal-infos
+        run: dart analyze --no-fatal-warnings
```
clintpurser, line 25: On default, `dart analyze` will fail if it finds `warnings`. `--no-fatal-warnings` causes the check to not fail, this ought to be removed eventually to enforce failures on warnings, but for now we have a few that need to be cleaned up.

I would even further prefer adding  `--fatal-infos` which enforces failures on the `info` level, it will take a bit more work to get there as well. In a few iterations I think we can be there.

.github/workflows/test.yaml
```diff
@@ -19,12 +19,14 @@ jobs:
             
       - run: flutter --version
       - run: flutter pub get
+      - run: flutter pub get --directory=example/viam_example_app
 
+      # Verify formatting in all of example, test & lib except lib/gen/
       - name: Verify formatting
-        run: dart format --output=none .
+        run: dart format --set-exit-if-changed example test lib/src/components lib/src/di lib/src/domain lib/src/media lib/src/proto lib/src/resource lib/src/robot lib/src/rpc
```
benjirewis, line 26: This seems like an inclusionary selection. It would be nice if it were exclusionary (just _ignored_ `lib/gen`). That way, we don't have to remember to add to this when new directories are added.

.github/workflows/test.yaml
benjirewis: SGTM; could optionally add a `TODO` here with ticket num for a ticket that removes all warnings and infos.

.vscode/settings.json
benjirewis: I've never seen a vscode settings file in our repo (not against the idea, just foreign to me). Do we need this file for CI? /Are we trying to support devs who use vscode having the same linting settings? I see you have `cmake.configureOnOpen: false` below which seems unrelated to dart.

.github/workflows/test.yaml
clintpurser, line 26: I wish there was a prettier way to do this. ☹️  perhaps we move where `gen/` is located? :thinking:

clintpurser: I agree, I don't think this is a good way to do it and i'd prefer to just exclude the one folder. I will look into that again 👀

.vscode/settings.json
clintpurser: The reason it's included in our flutter sdk is because VS Code is the supported IDE for Flutter/Dart, so there is a lot of tooling for it built in, so it makes sense to include it in the project so we can share settings for formatting etc. across people working on it.

the cmake is unrelated to dart, im not familiar with that or what it does in this context.

another setting to call out is the `"editor.formatOnSave"` which will format files to dart standard upon save, its handy for not accidentally committing bad format and failing tests, i prefer to enforce this enabled on projects to avoid easy mistakes and wasted CI resources.

benjirewis: That makes sense to me; thanks for the info!

.github/workflows/test.yaml
benjirewis: per offline discussion, if there's not a way to do this I wonder if we could do something like `dart format --set-exit-if-changed $(dart listpackages ./ | grep -v lib/gen)` where `list packages` is some command that may or may not exist to list all dart packages. no worries if not.

clintpurser: dart does not have a convenient way to get that list of files, i could write my own script, but that feels heavy handed. I would probably prefer to move the location of `gen/` before that.

.vscode/settings.json
njooma: Strongly against this. I'd prefer to keep the 140 line length as it makes everything SO MUCH easier to read, doesn't result in terrible indentations, etc. We can update the `dart format` command to include `--line-length=140`

We should also update the makefile with a `format` command that calls the same function `dart format --line-length=140`

.github/workflows/test.yaml
njooma: Use `find`: `dart format --line-length=140 $(find ./lib -name "*.dart" -not -path "./lib/src/gen/*")` will work on all files in `lib`, excluding files in `gen`

njooma (changes requested):

.vscode/settings.json
```diff
@@ -1,10 +1,4 @@
...
-    },
     "cmake.configureOnOpen": false,
```
njooma, line 2: This line can probably get nixed. I think I accidentally clicked on something that popped up in VSCode like "hey did you know there's a marketplace extension for this file" and i was like STOP

.github/workflows/test.yaml
clintpurser: great callout, that cleaned this up nicely. Thank you!!

.vscode/settings.json
clintpurser: 👋

.github/workflows/test.yaml
```diff
@@ -19,12 +19,14 @@ jobs:
             
       - run: flutter --version
       - run: flutter pub get
+      - run: flutter pub get --directory=example/viam_example_app
 
+      # Verify formatting in all dart files, except lib/gen/
       - name: Verify formatting
-        run: dart format --output=none .
+        run: dart format --line-length=140 --set-exit-if-changed $(find . -name "*.dart" -not -path "./lib/src/gen/*")
```
benjirewis, line 26: nice 🧑‍🔧

benjirewis (approved):

njooma (approved):

PR #29, opened 2023-05-24
RSDK-3124 - Board server side

clintpurser:
This PR adds the server side for Board.

Testing will be added in a follow up and is tracked here https://viam.atlassian.net/browse/RSDK-3283

### DIFF
--- a/lib/src/components/base/client.dart
+++ b/lib/src/components/base/client.dart
@@ -1,4 +1,4 @@
-import 'package:fixnum/src/int64.dart';
+import 'package:fixnum/fixnum.dart';
 import 'package:grpc/grpc_connection_interface.dart';
 import 'package:viam_sdk/src/gen/common/v1/common.pb.dart';
 import 'package:viam_sdk/src/gen/component/base/v1/base.pbgrpc.dart';
--- a/lib/src/components/board/board.dart
+++ b/lib/src/components/board/board.dart
@@ -1,9 +1,9 @@
-import 'package:fixnum/src/int64.dart';
-import 'package:viam_sdk/src/proto/common.dart';
-import 'package:viam_sdk/src/resource/base.dart';
+import 'package:fixnum/fixnum.dart';
 
 import '../../gen/common/v1/common.pb.dart' as common;
 import '../../gen/component/board/v1/board.pbenum.dart';
+import '../../proto/common.dart';
+import '../../resource/base.dart';
 import '../../robot/client.dart';
 
 class BoardStatus {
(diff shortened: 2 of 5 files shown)


clintpurser:
CI will fail till https://github.com/viamrobotics/viam-flutter-sdk/pull/27 is in.

lib/src/components/board/client.dart
```diff
@@ -15,6 +16,12 @@ class BoardClient extends Board {
 
   BoardClient(this.name, this._channel) : _client = BoardServiceClient(_channel);
 
+  @override
+  Future<Map<String, dynamic>> doCommand(Map<String, dynamic>? command) async {
+    final response = await _client.doCommand(common.DoCommandRequest(name: name, command: command?.toStruct()));
+    return response.result.toMap();
+  }
+
```
clintpurser, line 24: driveby: board didn't have doCommand implemented

lib/src/components/board/board.dart
```diff
@@ -16,6 +18,13 @@ class BoardStatus {
     pbBoardStatus.digitalInterrupts.forEach((key, value) => boardStatus.digitalInterrupts[key] = (value.value.toInt()));
     return boardStatus;
   }
+
+  common.BoardStatus toPbBoardStatus() {
```
njooma, line 22: I think this would be better as a ` get` attribute called `proto`

```suggestion
  common.BoardStatus get proto {
```

lib/src/components/board/board.dart
```diff
@@ -1,3 +1,5 @@
+import 'package:fixnum/src/int64.dart';
+import 'package:viam_sdk/src/proto/common.dart';
 import 'package:viam_sdk/src/resource/base.dart';
```
njooma, line 3: These should be relative imports

lib/src/components/board/client.dart
```diff
@@ -2,6 +2,7 @@ import 'package:fixnum/src/int64.dart';
 import 'package:grpc/grpc_connection_interface.dart';
 import 'package:viam_sdk/src/gen/google/protobuf/duration.pb.dart' as grpc_duration;
```
njooma, line 3: Relative import

njooma (changes requested):
Small changes

lib/src/components/board/board.dart
clintpurser: i like that, good call.

lib/src/components/board/board.dart
clintpurser: ah yeah, thanks. 👀

lib/src/components/base/client.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:fixnum/src/int64.dart';
+import 'package:fixnum/fixnum.dart';
 import 'package:grpc/grpc_connection_interface.dart';
 import 'package:viam_sdk/src/gen/common/v1/common.pb.dart';
 import 'package:viam_sdk/src/gen/component/base/v1/base.pbgrpc.dart';
```
njooma, line 4: Local imports :)

njooma (approved):
Nit on imports

lib/src/components/base/client.dart
clintpurser: noooo i thought i got these 😭

clintpurser: oh this is in base 👀

njooma: Oh that's on me then 😬

PR #30, opened 2023-05-24
RSDK-3303 - first cut camera server side

clintpurser:
This PR adds the server side for camera. 

Testing will be added in a follow up, that is being tracked in https://viam.atlassian.net/browse/RSDK-3304

### DIFF
--- a/lib/src/components/camera/service.dart
+++ b/lib/src/components/camera/service.dart
@@ -13,7 +13,7 @@ class CameraService extends CameraServiceBase {
 
   CameraService(this._manager);
 
-  Camera _cameraFromManager(String name) {
+  Camera _fromManager(String name) {
     try {
       return _manager.getResource(Camera.getResourceName(name));
     } catch (e) {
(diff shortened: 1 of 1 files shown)


lib/src/components/camera/service.dart
```diff
@@ -0,0 +1,60 @@
...
+        supportsPcd: result.supportsPcd,
+        intrinsicParameters: result.intrinsicParameters,
+        distortionParameters: result.distortionParameters);
+  }
+
+  @override
+  Future<HttpBody> renderFrame(ServiceCall call, RenderFrameRequest request) {
+    // TODO: implement renderFrame
+    throw UnimplementedError();
+  }
```
clintpurser, line 59: @njooma this method isn't in the client, I'm not sure what to do with it, included as Unimplemented for now.

clintpurser: oh duh i forgot to check other SDKs... 👀 looking at [python](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/components/camera/service.py#LL54C3-L73C44) then others...

clintpurser: ✅

clintpurser:
ci will fail till https://github.com/viamrobotics/viam-flutter-sdk/pull/27 is in, i will override while merging if this gets approved before.

lib/src/components/camera/service.dart
```diff
@@ -0,0 +1,62 @@
...
+    Camera camera = _cameraFromManager(request.name);
+    var result = await camera.doCommand(request.command.toMap());
+    return DoCommandResponse(result: result.toStruct());
+  }
+
+  @override
+  Future<GetImageResponse> getImage(ServiceCall call, GetImageRequest request) async {
+    Camera camera = _cameraFromManager(request.name);
+    var result = await camera.image(mimeType: MimeType.fromString(request.mimeType));
+    return GetImageResponse(mimeType: result.mimeType.toString(), image: result.raw);
```
clintpurser, line 35: I am unsure if `result.mimeType.toString()` will convert the mimeType like we want, I will verify when writing the tests.

njooma (approved):
LGTM. Can't wait for tests :P

lib/src/components/camera/service.dart
```diff
@@ -0,0 +1,61 @@
...
+        intrinsicParameters: properties.intrinsicParameters,
+        distortionParameters: properties.distortionParameters);
+  }
+
+  @override
+  Future<HttpBody> renderFrame(ServiceCall call, RenderFrameRequest request) async {
+    final camera = _fromManager(request.name);
+    final image = await camera.image(mimeType: MimeType.fromString(request.mimeType));
+    return HttpBody(data: image.raw, contentType: image.mimeType.toString());
+  }
```
maximpertsov, line 60: Q: are planning to convert our custom rbga images here or in the client (or just punting for now)

maximpertsov: CC @njooma since you might have more context here

clintpurser: I don't understand what you're asking.

I tried to copy what this method does in the python and cpp sdk for this method, but not doing any `converting` that i know of.

maximpertsov: so we have a custom image mime type called `image/vnd.viam.rgba`, that requires custom logic to decode (e.g. [here's the logic that does that in python](https://github.com/viamrobotics/viam-python-sdk/blob/7ae6a3cbc4f80295c61a4104a751a35d0d94836a/src/viam/media/viam_rgba_plugin.py))

however i think i know the answer to my own question - we need to decode it to display it properly so this logic will be applied client-side

maximpertsov (approved):

lib/src/components/camera/service.dart
njooma: @maximpertsov I already took care of ViamRGBA type [here](https://github.com/viamrobotics/viam-flutter-sdk/blob/1f8acb5a19984e6100b2b698ceaa051c45786aa9/lib/src/media/image.dart#L107). Everything is taken care of under the hood :)

PR #31, opened 2023-05-24
[RSDK-3126] Base Service + Tests

njooma:
Introduce tests for the Base resource, along with its service

### DIFF
--- a/lib/src/components/base/service.dart
+++ b/lib/src/components/base/service.dart
@@ -1,0 +1,70 @@
+import 'package:grpc/grpc.dart';
+
+import '../../gen/common/v1/common.pb.dart';
+import '../../gen/component/base/v1/base.pbgrpc.dart';
+import '../../resource/manager.dart';
+import '../../utils.dart';
+import 'base.dart';
+
+class BaseService extends BaseServiceBase {
+  final ResourceManager _manager;
+
+  BaseService(this._manager);
+
+  Base _fromManager(String name) {
+    try {
+      return _manager.getResource(Base.getResourceName(name));
+    } catch (e) {
+      throw GrpcError.notFound(e.toString());
+    }
+  }
+
+  @override
+  Future<MoveStraightResponse> moveStraight(ServiceCall call, MoveStraightRequest request) async {
+    Base base = _fromManager(request.name);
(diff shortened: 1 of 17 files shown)


lib/src/components/base/service.dart
```diff
@@ -0,0 +1,70 @@
...
+
+  BaseService(this._manager);
+
+  Base _fromManager(String name) {
+    try {
+      return _manager.getResource(Base.getResourceName(name));
+    } catch (e) {
+      throw GrpcError.notFound(e.toString());
+    }
+  }
```
clintpurser, line 20: I like how you named this, omitting `base`. I will copy in other services.

I've been thinking how we can abstract some of the boiler plate code thats essentially copied between each service, maybe a service `mixin` would be a good move to abstract this out and include there.

lib/src/components/servo/service.dart
```diff
@@ -11,45 +11,45 @@ class ServoService extends ServoServiceBase {
 
   ServoService(this._manager);
 
-  Servo _servoFromManager(String name) {
+  Servo _fromManager(String name) {
```
clintpurser, line 14: oh! you already got servo, nice. 🙏

test/unit_test/components/base_test.dart
```diff
@@ -0,0 +1,381 @@
...
+      expect(base.extra, {'foo': 'bar'});
+    });
+  });
+
+  group('Base RPC Tests', () {
+    late ClientChannel channel;
+    late FakeBase base;
+    late BaseService service;
+    late Server server;
+    late String name;
```
clintpurser, line 176: `nit` same as above
```suggestion
    const String name = 'base';
```

lib/src/components/base/service.dart
njooma: Ya same. Tried abstracting it and ran into issues around not having abstract static functions. Renamed to be just `_fromManager` to make it easier to copy and paste until we figure out a way to refactor it lol

clintpurser: SGTM, I'm down for finding the right abstractions after we have things fleshed out.

clintpurser (approved):
LGTM - a few nits that are super minor.

PR #35, opened 2023-05-24
RSDK-3127 - Movement sensor server

clintpurser:
testing will be added in a follow up. being tracked here [RSDK-3282](https://viam.atlassian.net/browse/RSDK-3282)

### DIFF
--- a/lib/src/components/movement_sensor/service.dart
+++ b/lib/src/components/movement_sensor/service.dart
@@ -1,0 +1,90 @@
+import 'package:grpc/grpc.dart';
+
+import '../../gen/common/v1/common.pb.dart';
+import '../../gen/component/movementSensor/v1/movementSensor.pbgrpc.dart';
+import '../../resource/manager.dart';
+import '../../utils.dart';
+import 'movement_sensor.dart';
+
+class MovementSensorService extends MovementSensorServiceBase {
+  final ResourceManager _manager;
+
+  MovementSensorService(this._manager);
+
+  MovementSensor _fromManager(String name) {
+    try {
+      return _manager.getResource(MovementSensor.getResourceName(name));
+    } catch (e) {
+      throw (GrpcError.notFound(e.toString()));
+    }
+  }
+
+  @override
+  Future<DoCommandResponse> doCommand(ServiceCall call, DoCommandRequest request) async {
+    final MovementSensor movementSensor = _fromManager(request.name);
+    final result = await movementSensor.doCommand(request.command.toMap());
+    return DoCommandResponse(result: result.toStruct());
+  }
+
+  @override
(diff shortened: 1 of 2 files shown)


lib/src/components/sensor/service.dart
```diff
@@ -11,7 +11,7 @@ class SensorService extends SensorServiceBase {
 
   SensorService(this._manager);
 
-  Sensor _sensorFromManager(String name) {
+  Sensor _fromManager(String name) {
```
clintpurser, line 14: diveby: noticed wanted to update this naming, and use finals in this file.

clintpurser:
I don't know why the CI is failing. I _think_ it might be a red hearing and we can ignore it.

https://github.com/viamrobotics/viam-flutter-sdk/pull/35/checks
```
error - lib/src/components/movement_sensor/service.dart:4:8 - Target of URI doesn't exist: '../../gen/component/movementSensor/v1/movementSensor.pbgrpc.dart'. Try creating the file referenced by the URI, or try using a URI for a file that does exist. - uri_does_not_exist
```

it's not seeing the pbgrpc file for some reason. I will investigate tomorrow. 👀

lib/src/components/movement_sensor/service.dart
```diff
@@ -0,0 +1,90 @@
...
+  Future<GetAccuracyResponse> getAccuracy(ServiceCall call, GetAccuracyRequest request) async {
+    final MovementSensor movementSensor = _fromManager(request.name);
+    final accuracyMm = await movementSensor.accuracy(extra: request.extra.toMap());
+    return GetAccuracyResponse(accuracyMm: accuracyMm);
+  }
+
+  @override
+  Future<GetAngularVelocityResponse> getAngularVelocity(ServiceCall call, GetAngularVelocityRequest request) async {
+    final MovementSensor movementSensor = _fromManager(request.name);
+    final Vector3 angularVelocity = await movementSensor.angularVelocity(extra: request.extra.toMap());
```
njooma, line 39: Q: why explicitly add the type here (and elsewhere in this file)?

clintpurser: For no good reason haha. I originally had them all defined without the type, but then thought it was more clear to add the type.

There was also one spot where the type inference didn't work in a different service because of an overloaded type name, but that isn't a reason to type everything strictly. 

I read over darts suggestions on annotating type, seems we ought to not annotate local variables unless we need to.

> Don’t annotate locals and generic invocations unless you need to.

https://dart.dev/effective-dart/design#types

I'll rework this, and then make sure other servers follow suit, i did this elsewhere.

clintpurser: https://github.com/viamrobotics/viam-flutter-sdk/pull/36

cheukt (approved):

PR #36, opened 2023-05-25
Prefer final locals

clintpurser:
added lint rule prefer_final_locals

> Declaring variables as final when possible is a good practice because it helps avoid accidental reassignments and allows the compiler to do optimizations.

https://dart-lang.github.io/linter/lints/prefer_final_locals.html

Also removed some unnecessary type annotations - https://dart.dev/effective-dart/design#types

> Don’t annotate locals and generic invocations unless you need to.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,4 +29,4 @@ jobs:
         run: dart analyze --no-fatal-warnings
 
       - name: Run tests
-        run: flutter test
+        run: flutter test --concurrency=1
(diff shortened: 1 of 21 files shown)


clintpurser:
I also considered adding [omit local variable types](https://dart-lang.github.io/linter/lints/prefer_final_locals.html) but that felt a bit heavy handed. I'll keep thinking about if we want that one as well.

cheukt (approved):

njooma (approved):

PR #40, opened 2023-05-31
RSDK-3281 - Sensor tests

clintpurser:
- Add Sensor tests (FakeSensor & Tests, Service Tests, Client Tests)
- add doCommand for sensor
- fix toValue() for List<T>

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -48,17 +48,40 @@ abstract class ValueType {
   Value toValue();
 }
 
-// TODO fix To toValue() for List<T?>
-extension ListValueUtils<T extends ValueType> on List<T> {
+extension ListValueUtils<T> on List<T> {
   Value toValue() {
-    return Value(listValue: ListValue(values: map((e) => e.toValue())));
+    final values = map((e) {
+      try {
+        if (e is num) {
+          return Value(numberValue: e.toDouble());
+        } else if (e is String) {
+          return Value(stringValue: e);
+        } else if (e is bool) {
+          return Value(boolValue: e);
+        } else if (e is List<dynamic>) {
+          return e.toValue();
+        } else if (e is Map<String, dynamic>) {
+          return e.toValue();
+          // ignore: type_check_with_null, prefer_void_to_null
+        } else if (e is Null) {
+          return Value(nullValue: e);
+        } else {
+          throw GrpcError.invalidArgument('List contains unsupported type');
+        }
+      } catch (error, st) {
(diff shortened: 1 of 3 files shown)


lib/src/utils.dart
```diff
@@ -48,17 +48,40 @@ abstract class ValueType {
...
+
+    return Value(listValue: ListValue(values: values));
   }
 }
 
 extension MapStructUtils on Map<String, dynamic> {
   Struct toStruct() {
     final Map<String, Value> result = {};
-    for (var entry in entries) {
+    for (final entry in entries) {
```
clintpurser, line 84: driveby, this ought to be final.

lib/src/utils.dart
```diff
@@ -48,17 +48,40 @@ abstract class ValueType {
...
+          return Value(boolValue: e);
+        } else if (e is List<dynamic>) {
+          return e.toValue();
+        } else if (e is Map<String, dynamic>) {
+          return e.toValue();
+          // ignore: type_check_with_null, prefer_void_to_null
+        } else if (e is Null) {
+          return Value(nullValue: e);
+        } else {
+          throw GrpcError.invalidArgument('List contains unsupported type');
```
clintpurser, line 69: I attempted to extract this logic into a helper function that both this and  Map<> to.Value() could use, because it is essentially copied from there, but I started running into the same issues we had before where it didn't find the correct toValue() at runtime, I may take a look at this again with a fresh mind.

clintpurser:
😭 more test failures from tests occupying the same port... 👀

test/unit_test/components/gantry_test.dart
```diff
@@ -119,9 +119,9 @@ void main() {
     late Server server;
 
     setUp(() async {
-      final port = 50000 + (StackTrace.current.hashCode % 10000);
+      final port = 50000 + (name.hashCode % 10000);
```
cheukt, line 122: seems like a good function to pull out into some kind of a testutils file

cheukt (approved):
small suggestion but otherwise lgtm

cheukt (approved):
small suggestion but otherwise lgtm

test/unit_test/components/gantry_test.dart
cheukt: sgtm

clintpurser: https://github.com/viamrobotics/viam-flutter-sdk/pull/41 includes a helper function

njooma (approved):

PR #41, opened 2023-06-01
RSDK-3283 - Board tests

clintpurser:
- FakeBoard
- Board Tests (FakeBoard, Server, Client)
- misc Board Client fixes the tests helped identify
- testing port helper function

### DIFF
--- a/lib/src/components/board/board.dart
+++ b/lib/src/components/board/board.dart
@@ -13,10 +13,8 @@ class BoardStatus {
   const BoardStatus(this.analogs, this.digitalInterrupts);
 
   factory BoardStatus.fromProto(common.BoardStatus pbBoardStatus) {
-    final Map<String, int> analogs = {};
-    final Map<String, int> digitalInterrupts = {};
-    pbBoardStatus.analogs.forEach((key, value) => analogs[key] = value.value);
-    pbBoardStatus.digitalInterrupts.forEach((key, value) => digitalInterrupts[key] = (value.value.toInt()));
+    final Map<String, int> analogs = pbBoardStatus.analogs.map((key, value) => MapEntry(key, value.value));
+    final Map<String, int> digitalInterrupts = pbBoardStatus.digitalInterrupts.map((key, value) => MapEntry(key, value.value.toInt()));
     return BoardStatus(analogs, digitalInterrupts);
   }
 
--- a/lib/src/components/board/board.dart
+++ b/lib/src/components/board/board.dart
@@ -13,10 +13,11 @@ class BoardStatus {
   const BoardStatus(this.analogs, this.digitalInterrupts);
 
   factory BoardStatus.fromProto(common.BoardStatus pbBoardStatus) {
-    const boardStatus = BoardStatus(<String, int>{}, <String, int>{});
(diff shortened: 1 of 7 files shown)


test/test_utils.dart
```diff
@@ -0,0 +1 @@
+int generateTestingPortFromName(String name) => 50000 + (name.hashCode % 10000);
```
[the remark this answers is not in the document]
maximpertsov: do we actually need this function? why not just hardcode a port per test?

maximpertsov (approved):

test/test_utils.dart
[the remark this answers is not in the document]
clintpurser: we tossed that around as an idea, I prefer an approach like this because we will have a lot of tests and manually tracking which ports have been used could be prone to errors.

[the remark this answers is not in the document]
maximpertsov: cool, as for the name - nothing else comes to mind - maybe it can be shorted to just `generateTestingPort` for brevity's sake?

```suggestion
int generateTestingPort(String name) => 50000 + (name.hashCode % 10000);
```

lib/src/components/board/board.dart
```diff
@@ -13,10 +13,11 @@ class BoardStatus {
...
 
   factory BoardStatus.fromProto(common.BoardStatus pbBoardStatus) {
-    const boardStatus = BoardStatus(<String, int>{}, <String, int>{});
-    pbBoardStatus.analogs.forEach((key, value) => boardStatus.analogs[key] = value.value);
-    pbBoardStatus.digitalInterrupts.forEach((key, value) => boardStatus.digitalInterrupts[key] = (value.value.toInt()));
-    return boardStatus;
+    final Map<String, int> analogs = {};
+    final Map<String, int> digitalInterrupts = {};
+    pbBoardStatus.analogs.forEach((key, value) => analogs[key] = value.value);
+    pbBoardStatus.digitalInterrupts.forEach((key, value) => digitalInterrupts[key] = (value.value.toInt()));
```
njooma, line 19: This is a little cleaner IMO. Thoughts? 
```suggestion
    final Map<String, int> analogs = pbBoardStatus.analogs.map((key, value) => MapEntry(key, value.value));
    final Map<String, int> digitalInterrupts = pbBoardStatus.digitalInterrupts.map((key, value) => MapEntry(key, value.value.toInt()));
```

lib/src/components/board/client.dart
```diff
@@ -31,17 +31,18 @@ class BoardClient extends Board {
   @override
   Future<void> setGpioState(String pin, bool high, {Map<String, dynamic>? extra}) async {
     await _client.setGPIO(SetGPIORequest(name: name, pin: pin, high: high, extra: extra?.toStruct()));
+    return;
```
njooma, line 34: Why is this needed?

clintpurser: oh good catch thank you! it is not needed. I added that while debugging and did not mean to include it. 🪓

lib/src/components/board/board.dart
clintpurser: yeah I like that, thank you!

njooma (approved):

PR #42, opened 2023-06-02
RSDK-3128 - Remote control (Base) example

clintpurser:
- adds logo for example app
- adds base (rc) screen
- widgetizes stream widget and base joystick widget.

I don't love the way i'm passing in a camera and stream and base into the widgets / new screens. I think passing around a RescourceName would be cleaner, that will require a little bit of refactoring which I want to do in a separate follow up PR, as to not spend too much time on this right now.


example/viam_example_app/lib/main.dart
```diff
@@ -74,6 +75,7 @@ class MyHomePage extends StatefulWidget {
 class _MyHomePageState extends State<MyHomePage> {
   bool _loggedIn = false;
   bool _loading = false;
+  late ResourceName? _cameraName;
```
njooma, line 78: This should either be `late` or `optional`, no need for it to be both

njooma (approved):

example/viam_example_app/lib/main.dart
clintpurser: good catch, thank you!

.DS_Store
cheukt, line 1: delete?

cheukt (approved):

.DS_Store
clintpurser: 👀 huh - yeah lets axe this

njooma: Added it here: https://github.com/viamrobotics/viam-flutter-sdk/pull/45/commits/7f462a98744bad564f9daf8560a7c03c1a01f4ad

example/viam_example_app/.metadata
```diff
@@ -13,26 +13,11 @@ project_type: app
...
-      create_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
-      base_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
-    - platform: web
-      create_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
-      base_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
-    - platform: windows
-      create_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
-      base_revision: 90c64ed42ba53a52d18f0cb3b17666c8662ed2a0
+      create_revision: 84a1e904f44f9b0e9c4510138010edcc653163f8
+      base_revision: 84a1e904f44f9b0e9c4510138010edcc653163f8
```
clintpurser, line 20: i think we can remove this file, i'll do so in a follow up. probably lots to clean up in the example app.

PR #44, opened 2023-06-05
RSDK-3304 - camera tests

clintpurser:
- add doCommand
- add FakeCamera
- Tests for FakeCamera, Client and service
- fix mimetype tostring

### DIFF
--- a/lib/src/components/camera/service.dart
+++ b/lib/src/components/camera/service.dart
@@ -32,14 +32,14 @@ class CameraService extends CameraServiceBase {
   Future<GetImageResponse> getImage(ServiceCall call, GetImageRequest request) async {
     final camera = _fromManager(request.name);
     final image = await camera.image(mimeType: MimeType.fromString(request.mimeType));
-    return GetImageResponse(mimeType: image.mimeType.toString(), image: image.raw);
+    return GetImageResponse(mimeType: image.mimeType.name, image: image.raw);
   }
 
   @override
   Future<GetPointCloudResponse> getPointCloud(ServiceCall call, GetPointCloudRequest request) async {
     final camera = _fromManager(request.name);
     final image = await camera.pointCloud();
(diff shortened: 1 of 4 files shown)


lib/src/components/camera/service.dart
```diff
@@ -32,14 +32,14 @@ class CameraService extends CameraServiceBase {
...
-    return GetImageResponse(mimeType: image.mimeType.toString(), image: image.raw);
+    return GetImageResponse(mimeType: image.mimeType.name, image: image.raw);
   }
 
   @override
   Future<GetPointCloudResponse> getPointCloud(ServiceCall call, GetPointCloudRequest request) async {
     final camera = _fromManager(request.name);
     final image = await camera.pointCloud();
-    return GetPointCloudResponse(mimeType: image.mimeType.toString(), pointCloud: image.raw);
+    return GetPointCloudResponse(mimeType: image.mimeType.name, pointCloud: image.raw);
```
clintpurser, line 42: this was a bug i introduced while writing this service, now that we have tests I know for sure this needs to be changed. `name` is the value i want, `toString()` returned `"instance of \MimeType\"`.

njooma: [feat] maybe override the `toString` method on MimeType to return the name?

clintpurser: I considered that, i think grabbing `name` is fine for what we are expecting. overriding toString _might_ be overkill for this, open to it though. It would be more appropriate if we wanted to combine multiple fields from the type in a certain way IMO.

njooma: I actually think it's super useful to override the `toString`, as it would allow people to debug log/print the mimetypes easier if they're making their own camera implementation. `print("Received mimetype: $mimetype")` vs `print("Received mimetype: ${mimetype.name}") which tbf isn't a lot, but feels more ergonomic IMO

njooma (approved):

cheukt (approved):

PR #46, opened 2023-06-06
RSDK-3282 - Refactor channel

clintpurser:
refactoring `_channel` to no longer be a member of our clients, because it is only used inside the constructor to get `_client`. This resolves the error we were getting:
```
The value of the field '_channel' isn't used.
Try removing the field, or using it.
```
this allows us to enable more thorough static analysis on our CI and improve code quality.

I might not be understanding the purpose of `_channel` and if we really do need further access to it in our clients, would love to know what im missing if so!


clintpurser:
Thanks for explaining @njooma - I agree in that case I'd prefer to hold off on this change, since that reconnection work is in our current slate of work. Thanks for the context!! Closing...

PR #50, opened 2023-06-07
RSDK-3387 - add utils tests

clintpurser:
adds tests for utils and test utils. refactors utils a bit.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -17,23 +17,19 @@ extension NullableStringUtils on String? {
 
 extension ValueUtils on Value {
   dynamic toPrimitive() {
-    if (hasNullValue()) {
-      return nullValue;
-    }
     if (hasNumberValue()) {
       return numberValue;
-    }
-    if (hasStringValue()) {
+    } else if (hasStringValue()) {
       return stringValue;
-    }
-    if (hasBoolValue()) {
+    } else if (hasBoolValue()) {
       return boolValue;
-    }
-    if (hasStructValue()) {
+    } else if (hasStructValue()) {
       return structValue.fields.map((key, value) => MapEntry(key, value.toPrimitive()));
-    }
-    if (hasListValue()) {
+    } else if (hasListValue()) {
       return listValue.values.map((e) => e.toPrimitive());
+    } else {
+      // default to null if no value found
+      return null;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


lib/src/utils.dart
```diff
@@ -17,23 +17,19 @@ extension NullableStringUtils on String? {
...
-    }
-    if (hasStructValue()) {
+    } else if (hasStructValue()) {
       return structValue.fields.map((key, value) => MapEntry(key, value.toPrimitive()));
-    }
-    if (hasListValue()) {
+    } else if (hasListValue()) {
       return listValue.values.map((e) => e.toPrimitive());
+    } else {
+      // default to null if no value found
```
clintpurser, line 31: @njooma I realized if there was no value we wouldn't return anything, do we want to default to null?

stuqdog (approved):
LGTM % question

lib/src/utils.dart
```diff
@@ -17,23 +17,19 @@ extension NullableStringUtils on String? {
...
 extension ValueUtils on Value {
   dynamic toPrimitive() {
-    if (hasNullValue()) {
-      return nullValue;
-    }
     if (hasNumberValue()) {
       return numberValue;
-    }
-    if (hasStringValue()) {
+    } else if (hasStringValue()) {
```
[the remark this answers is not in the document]
clintpurser: Great question, this was something I was unsure about as well and I'm trying to get clarity on in this PR review process.

 I am unsure if multiple of the Value() checks _should_ return true, it seems like they should not. But there is nothing in the code AFAIK enforcing a Value from not having multiple value types inside it. 

Even before shifting to `else if` it would have just returned whatever one it found first. The ordering was arbitrary from what I understand, it still is arbitrary other than moving `null` to be last as I'm proposing it be the default.

lib/src/utils.dart
njooma: It should always have one of those fields set. So I don’t think null is necessarily bad and it is defensive, but I feel like an error might also make sense because it should be impossible to get to that state. But null is fine probs

lib/src/utils.dart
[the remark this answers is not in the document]
clintpurser: ooh awesome, that definition is super helpful, thank you!

lib/src/utils.dart
clintpurser: I think you're right, throwing an error makes a lot of sense. I will rework this.

PR #52, opened 2023-06-22
Removing Podfile.lock since it causes issues without a pubspec.lock

njooma:
When running the example app, we often have to reset the cocoapods packages for ios/macos because of version mismatches. Removing the podfile.lock by default should help with those issues


clintpurser (approved):
good catch! LGTM

PR #54, opened 2023-06-27
RSDK-3192 - Add robot reconnection logic

clintpurser:
This PR adds robot reconnection logic, I copied the logic from our python SDK, so it ought to be very similar to that.

There are a couple of fly by's included here, i will call them out below.

Some follow up work will be reworking our stream client to reconnect as well. this is under construction currently in [this](https://github.com/viamrobotics/viam-flutter-sdk/pull/53) - i will address in a follow up.

### DIFF
--- a/lib/src/resource/base.dart
+++ b/lib/src/resource/base.dart
@@ -1,3 +1,6 @@
+import 'package:grpc/grpc.dart';
+import 'package:grpc/grpc_connection_interface.dart';
+
 import '../gen/common/v1/common.pb.dart';
 
 const String resourceNamespaceRDK = 'rdk';
@@ -35,4 +38,10 @@ abstract class Resource {
   Future<Map<String, dynamic>> doCommand(Map<String, dynamic> command) {
     throw UnimplementedError();
   }
+}
+
+abstract class ResourceRPCClient<T extends Client> {
+  abstract ClientChannelBase channel;
+
+  T get client;
 }
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -21,7 +21,7 @@ class RobotClientOptions {
   final DialOptions dialOptions;
 
   /// The frequency (in seconds) at which to check if the robot is still connected. 0 (zero) signifies no connection checks
-  final checkConnectionInterval = 1;
+  final checkConnectionInterval = 10;
 
   /// The frequency (in seconds) at which to attempt to reconnect a disconnected robot. 0 (zero) signifies no reconnection attempts
   final attemptReconnectInterval = 1;
@@ -59,7 +59,7 @@ class RobotClient {
     client._options = options.dialOptions;
     client._channel = await dial(url, options.dialOptions);
     client._client = RobotServiceClient(client._channel);
-    client._streamManager = StreamManager(client.channel as WebRtcClientChannel);
+    client._streamManager = StreamManager(client._channel as WebRtcClientChannel);
     await client.refresh();
     unawaited(client._checkConnection(interval: options.checkConnectionInterval, reconnectInterval: options.attemptReconnectInterval));
     return client;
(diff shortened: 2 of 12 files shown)


lib/src/resource/base.dart
```diff
@@ -36,3 +39,9 @@ abstract class Resource {
     throw UnimplementedError();
   }
 }
+
+abstract class ResourceRPCClient<T extends Client> {
+  abstract ClientChannelBase channel;
+
+  T get client;
+}
```
clintpurser, line 47: this abstract class helps us enforce that each component has a channel and client.

lib/src/components/movement_sensor/movement_sensor.dart
```diff
@@ -52,54 +52,7 @@ abstract class MovementSensor extends Sensor {
...
-
-    try {
-      final orient = await orientation(extra: extra);
-      readings['orientation'] = orient;
-    } catch (exception) {
-      // TODO: Check if the exception is of a specific type and ignore or rethrow
-    }
-
-    return readings;
-  }
```
clintpurser, line 102: i _dont think_ this logic should have been here, moved to the client where i think it makes more sense with our current patterns.

test/unit_test/components/movement_sensor_test.dart
```diff
@@ -267,6 +277,24 @@ void main() {
...
+          'linear_acceleration',
+          'compass',
+          'orientation'
+        ];
+        expect(result.keys, expectedKeys);
+
+        final expectedValues = [GeoPoint(latitude: 0.0), 0.0, Vector3(x: 0.0), Vector3(x: 0.0), Vector3(x: 0.0), 0.0, Orientation(oX: 0.0)];
+        expect(result.values, expectedValues);
+      });
+
```
clintpurser, line 297: needed to add missing tests for `readings` now that its in the client.

lib/src/components/movement_sensor/movement_sensor.dart
njooma: I think it should stay here, that way all subclasses of this resource will have this implementation out of the box (which I think is a good default)

njooma (changes requested):

lib/src/robot/client.dart
```diff
@@ -82,6 +102,63 @@ class RobotClient {
...
+          await _client.resourceNames(ResourceNamesRequest(), options: CallOptions(timeout: const Duration(seconds: 1)));
+          _connected = true;
+          break;
+        } catch (e) {
+          _connected = false;
+          await Future.delayed(const Duration(milliseconds: 100));
+        }
+      }
+      if (_connected) {
+        _logger.d('Robot still connected ${DateTime.now()}');
```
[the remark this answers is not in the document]
clintpurser: that makes sense to me, I will remove this.

lib/src/components/movement_sensor/movement_sensor.dart
clintpurser: SGTM I will move this back.

lib/src/robot/client.dart
```diff
@@ -10,11 +13,19 @@ import '../resource/registry.dart';
...
 
+Logger _logger = Logger();
+
 /// The options that define the behavior of the [RobotClient].
 class RobotClientOptions {
   /// Options for connecting to the robot
   final DialOptions dialOptions;
 
+  /// The frequency (in seconds) at which to check if the robot is still connected. 0 (zero) signifies no connection checks
+  final checkConnectionInterval = 1;
```
cheukt, line 24: I believe in other sdks the default is 10s

cheukt (approved):
LGTM, slightly out of scope here but can we add a simple `Connected` method so users can find if the robot is connected?

lib/src/robot/client.dart
clintpurser: ooh you are correct, i changed this to 1 for debugging purposes. Good catch, thank you!

clintpurser:
@cheukt yeah I will add a `connected` method to the robot client... 👀

njooma (approved):

PR #56, opened 2023-06-29
Retrofit ViamImpl for access token auth

njooma:
Enable users to create a `Viam` instance with an access token only. This enables them to use app apis (getting organizations, data apis, etc.)


lib/src/viam_sdk_impl.dart
```diff
@@ -18,9 +24,37 @@ class ViamImpl implements Viam {
   ViamMovementService? movementService;
   ViamSensorService? sensorService;
   DataService? _dataService;
+  late AppClient _appClient;
```
stuqdog, line 27: (q) why are these `late`, but the above services are `null`able?

stuqdog (changes requested):

lib/src/viam_sdk_impl.dart
```diff
@@ -18,9 +24,37 @@ class ViamImpl implements Viam {
...
+  late DataClient _dataClient;
+
+  ViamImpl();
```
clintpurser, line 30:
```suggestion
  ViamImpl._();
```

can we make the default constructor private?

lib/src/viam_sdk_impl.dart
```diff
@@ -18,9 +24,37 @@ class ViamImpl implements Viam {
...
+
+  ViamImpl.withAccessToken(String accessToken) : _clientChannelBase = AuthenticatedChannel("app.viam.com", 443, accessToken, false) {
```
clintpurser, line 32:
```suggestion
  ViamImpl.withAccessToken(String accessToken) : _clientChannelBase = AuthenticatedChannel('app.viam.com', 443, accessToken, false) {
```

prefer single quotes

lib/src/viam_sdk.dart
```diff
@@ -1,29 +1,47 @@
...
+    return ViamImpl.withAccessToken(accessToken);
+  }
+
+  AppClient get appClient;
+  DataClient get dataClient;
+
+  Future<RobotClient> getRobotClient(Robot robot);
+
+  @Deprecated('This function will be removed prior to beta launch')
   static Viam instance() => ViamImpl();
```
clintpurser, line 27: do we want to give access to the object without an access token constructor?

clintpurser (changes requested):
overall lgtm - a few minor feedbacks, then ready to go.

lib/src/app/data.dart
```diff
@@ -0,0 +1,57 @@
...
+import '../domain/data/models/viam_data_request.dart';
+import '../domain/data/models/viam_tabular_data_response.dart';
+import '../gen/app/data/v1/data.pbgrpc.dart';
+
+class DataClient {
+  final DataServiceClient _client;
+
+  DataClient(ClientChannelBase channel) : _client = DataServiceClient(channel);
+
+  Future<ViamTabularDataResponse> tabularDataByFilter({
```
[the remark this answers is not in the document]
njooma: This is just a copy of the existing data APIs made by Droids on Roids. I'm not going to be changing any of this until later as it's out of scope.

lib/src/viam_sdk.dart
```diff
@@ -1,29 +1,47 @@
...
+  static Viam withAccessToken(String accessToken) {
+    return ViamImpl.withAccessToken(accessToken);
+  }
+
+  AppClient get appClient;
+  DataClient get dataClient;
+
+  Future<RobotClient> getRobotClient(Robot robot);
+
+  @Deprecated('This function will be removed prior to beta launch')
```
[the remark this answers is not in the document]
njooma: These deprecated functions/attributes were all provided by Droids on Roids. We will be removing them once we have our own system ready to go (which this PR starts on with `AppClient` and `DataClient`)

lib/src/viam_sdk_impl.dart
njooma: The ones I added are `late`. The ones DoR added are optional because that's how they set it up

lib/src/viam_sdk_impl.dart
njooma: We can't -- the (newly) deprecated version needs a public constructor and I don't want to mess with anything DoR might be using

lib/src/viam_sdk.dart
njooma: Yes -- because DoR might still be using it. I'm not sure where they are in their migration so I don't want to remove it (see above not about private constructor for `ViamImpl`)

njooma:
@stuqdog some context you might be missing -- DoR wrote a lot of this code. I'm simply doing some movement/migration so that it's flows with the rest of our SDK, but not making any big changes to functionality.

stuqdog:
> @stuqdog some context you might be missing -- DoR wrote a lot of this code. I'm simply doing some movement/migration so that it's flows with the rest of our SDK, but not making any big changes to functionality.

Ahh that makes sense, thanks for clarifying!

stuqdog (approved):

PR #57, opened 2023-07-03
RSDK-3861 - add generic client wrapper

clintpurser:
add a client wrapper for the `generic` component

### DIFF
--- a/lib/src/components/generic/client.dart
+++ b/lib/src/components/generic/client.dart
@@ -1,0 +1,26 @@
+import 'package:grpc/grpc_connection_interface.dart';
+
+import '../../gen/common/v1/common.pb.dart';
+import '../../gen/component/generic/v1/generic.pbgrpc.dart';
+import '../../resource/base.dart';
+import '../../utils.dart';
+import 'generic.dart';
+
+class GenericClient extends Generic implements ResourceRPCClient {
+  @override
+  ClientChannelBase channel;
+
+  @override
+  String name;
+
+  @override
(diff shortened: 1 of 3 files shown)


lib/src/components/generic/client.dart
```diff
@@ -0,0 +1,26 @@
...
+
+class GenericClient extends Generic implements ResourceRPCClient {
+  @override
+  ClientChannelBase channel;
+
+  @override
+  String name;
+
+  @override
+  GenericServiceClient get client => GenericServiceClient(channel);
```
stuqdog, line 17: (q) this looks to me like a constructor, but is using the syntax of a getter. Am I misunderstanding the use of `get`? Is this a common thing in dart?

stuqdog: (spoke with @clintpurser offline, this all makes sense!)

stuqdog (approved):

lib/src/components/generic/client.dart
clintpurser: Ethan and I talked offline about this. This is getting an instance of `GenericServiceClient` not `GenericClient`. We are doing it this way to not have to update the GenericServiceClient in each client upon the channel changing in the instance of a reconnection.

clintpurser:
this ended up not being needed for the demo app i am working on. but its done, so might as well carry on and merge it. 🤷

njooma (changes requested):
Can we get some tests plz?

clintpurser:
closing because this isn't a priority right now

PR #61, opened 2023-07-06
Update protos

njooma:
Lots of updates to proto, including the compiler which introduced a major breaking change around constructors (see [here](https://github.com/google/protobuf.dart/pull/703/files)).

This PR has a lot of generated files and a lot of the same type of change (removing args from constructor and moving them to cascading attrs), so feel free to rubber stamp. If you really want to look at what happened, I would suggest going commit by commit.


clintpurser (approved):
this is such an ugly change, im so sad about these API changes 😢 ... yolo 🚢

PR #65, opened 2023-07-13
[RSDK 3804] Fix webrtc streaming

njooma:
There was a bug where navigating to a camera stream page, then leaving, then coming back wouldn't properly show the stream. 

Turns out, this was due to an extremely subtle bug caused by the way we were originally dialing. Long story short, we were creating offers that contradicted each other, putting the webrtc streams in a mismatched state. The streaming server saw this weird state and set the client to inactive, so we never got updated when a new stream got added (for the second navigation back to camera).


clintpurser (approved):
nice!!

PR #66, opened 2023-07-14
RSDK-3057 - Sessions

clintpurser:
This PR adds sessions to the flutter SDK.

Testing Evidence:

https://github.com/viamrobotics/viam-flutter-sdk/assets/22106496/211ca662-2dd6-462d-a1cb-8c36cff225da

### DIFF
--- a/lib/src/di/di_clients/di_web_rtc_client.dart
+++ b/lib/src/di/di_clients/di_web_rtc_client.dart
@@ -15,8 +15,5 @@ Future<WebRtcClientChannel> _getWebRtcClient(
   final webRtcPeerConnection = WebRtcPeerConnection(webRtcDirectDataSource);
   await webRtcPeerConnection.createConnection();
 
-  return WebRtcClientChannel(
-    webRtcPeerConnection.peerConnection,
-    webRtcPeerConnection.dataChannel,
-  );
+  return WebRtcClientChannel(webRtcPeerConnection.peerConnection, webRtcPeerConnection.dataChannel, () => '');
 }
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:grpc/grpc_connection_interface.dart';
 import 'package:logger/logger.dart';
+import 'package:viam_sdk/src/robot/sessions_client.dart';
 
 import '../domain/web_rtc/web_rtc_client/web_rtc_client.dart';
 import '../gen/common/v1/common.pb.dart';
@@ -26,6 +27,9 @@ class RobotClientOptions {
   /// The frequency (in seconds) at which to attempt to reconnect a disconnected robot. 0 (zero) signifies no reconnection attempts
   final attemptReconnectInterval = 1;
 
+  /// Whether sessions are enabled
+  final enableSessions = true;
+
   RobotClientOptions() : dialOptions = DialOptions();
 
   /// Convenience initializer for creating options with specified [DialOptions]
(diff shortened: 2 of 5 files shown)


lib/src/robot/client.dart
```diff
@@ -26,6 +27,9 @@ class RobotClientOptions {
   /// The frequency (in seconds) at which to attempt to reconnect a disconnected robot. 0 (zero) signifies no reconnection attempts
   final attemptReconnectInterval = 1;
 
+  /// Whether sessions are enabled
+  final enableSessions = true;
```
stuqdog, line 31: (q) Is this customizable at all? It looks like this is always on by default. Are there cases where we want to support _not_ having sessions? If not, would it be better to just have it always-on and not customizable?

lib/src/robot/sessions_client.dart
```diff
@@ -0,0 +1,98 @@
...
+
+import '../resource/base.dart';
+
+final _logger = Logger();
+
+/// A Session allows a client to express that it is actively connected
+/// and supports stopping actuating components when it's not.
+class SessionsClient implements ResourceRPCClient {
+  static const sessionMetadataKey = 'viam-sid';
+  static const unallowedMethods = [
```
stuqdog, line 15: (minor/nit) I think making this a `set` would be more performant given how it's used?

lib/src/robot/client.dart
clintpurser: Great question. It is customizable, in the [docs for session](https://pkg.go.dev/go.viam.com/rdk/session) it is described as:

> Once implemented in an SDK, session management is opt-out from the client usage perspective.

Typescript and Python handle this similarly.

lib/src/robot/sessions_client.dart
clintpurser: Thats a great call out 👀 I will look into this. Thank you!

lib/src/robot/client.dart
stuqdog: Got it, thanks!

lib/src/di/di_clients/di_web_rtc_client.dart
```diff
@@ -15,8 +15,5 @@ Future<WebRtcClientChannel> _getWebRtcClient(
   final webRtcPeerConnection = WebRtcPeerConnection(webRtcDirectDataSource);
   await webRtcPeerConnection.createConnection();
 
-  return WebRtcClientChannel(
-    webRtcPeerConnection.peerConnection,
-    webRtcPeerConnection.dataChannel,
-  );
+  return WebRtcClientChannel(webRtcPeerConnection.peerConnection, webRtcPeerConnection.dataChannel, () => '');
```
benjirewis, line 18: Why the empty function here? WebRtcClientChannel's returned here have no associated session ID?

lib/src/robot/client.dart
```diff
@@ -145,19 +151,22 @@ class RobotClient {
...
         _channel = channel;
         _streamManager.channel = _channel as WebRtcClientChannel;
         _client = client;
+        _sessionsClient = SessionsClient(_channel, _options.enableSessions);
         await refresh();
         _connected = true;
         _logger.d('Successfully reconnected robot');
       } catch (e) {
         await _channel.shutdown();
+        _sessionsClient.reset();
```
benjirewis, line 169: Is this reset needed if we're resetting at the top of the reconnect loop?

benjirewis (approved):
LGTM; just some Qs!

lib/src/di/di_clients/di_web_rtc_client.dart
clintpurser: this is old Droids on roids code, naveed is actually deleting it here https://github.com/viamrobotics/viam-flutter-sdk/pull/68/files - so i just put this in as a placeholder, i should have called it out.

lib/src/robot/client.dart
clintpurser: it may not be, i copied where reset was called in python / TS, so it may be unneeded.

lib/src/di/di_clients/di_web_rtc_client.dart
benjirewis: Nice thanks.

PR #69, opened 2023-07-24
Add a tailLogs method

njooma:


lib/src/viam_sdk_impl.dart
```diff
@@ -13,7 +13,7 @@ class ViamImpl implements Viam {
   late AppClient _appClient;
   late DataClient _dataClient;
 
-  ViamImpl();
+  ViamImpl._();
```
clintpurser, line 16: just wondering, why make this constructor private?

clintpurser (approved):

PR #70, opened 2023-07-25
misc cleanups

clintpurser:
fixing some misc things that have come up in the SDK / example app.

1. the stream view would crash the app upon leaving it. changed it to use our cameraStreamViewWidget, it no longer crashes.

Before:

Screenshot 2023-07-25 at 10 30 41 AM

After:

https://github.com/viamrobotics/viam-flutter-sdk/assets/22106496/0ee4ac19-7cf5-4384-a292-323d1f04febf

2. GRPC was upgraded to 3.2.3, I was using version 3.2.0 locally and it was causing compile errors because of our const constructors on GRPC errors.

3. bumped the session heartbeat time down to 1.8x the default from 5x. we it was failing too often with that high amount of calls.

4. started a new session upon failing.

### DIFF
--- a/example/viam_example_app/lib/screens/stream.dart
+++ b/example/viam_example_app/lib/screens/stream.dart
@@ -1,10 +1,8 @@
-import 'dart:async' as async;
 import 'dart:typed_data';
 import 'dart:ui' as ui;
 
 import 'package:flutter/material.dart';
 import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
-import 'package:flutter_webrtc/flutter_webrtc.dart';
 import 'package:image/image.dart' as img;
 import 'package:viam_sdk/viam_sdk.dart';
 
@@ -22,12 +20,6 @@ class StreamScreen extends StatefulWidget {
 }
 
 class _StreamScreenState extends State<StreamScreen> {
-  // Stream
-  bool _ready = false;
-  bool _loading = false;
-  late RTCVideoRenderer _renderer;
-  late async.StreamSubscription<MediaStream> _streamSub;
-
   // Single frame
   ByteData? imageBytes;
   bool _imgLoaded = false;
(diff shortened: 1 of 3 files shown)


lib/src/robot/sessions_client.dart
```diff
@@ -55,8 +55,8 @@ class SessionsClient implements ResourceRPCClient {
         // We send heartbeats slightly faster than the interval window to
         // ensure that we don't fall outside of it and expire the session.
         _heartbeatInterval = Duration(
-          seconds: response.heartbeatWindow.seconds.toInt() ~/ 5,
-          microseconds: response.heartbeatWindow.nanos ~/ 5,
+          seconds: response.heartbeatWindow.seconds.toInt() ~/ 1.8,
```
jckras, line 58: How did you land on 1.8?

example/viam_example_app/lib/screens/stream.dart
```diff
@@ -73,35 +65,6 @@ class _StreamScreenState extends State<StreamScreen> {
     return uiImage;
   }
 
-  Future<void> _startStream() async {
```
jckras, line 76: why is this all being deleted?

lib/src/robot/sessions_client.dart
clintpurser: good question, i was playing with 2, and wanted slightly slower than 2. I don't think 1.8 makes any large difference from 2 and probably gets lost most of the time staying as an int. Basically, there is not hard and fast reason why i landed on 1.8

lib/src/robot/sessions_client.dart
```diff
@@ -65,6 +65,7 @@ class SessionsClient implements ResourceRPCClient {
       });
     } catch (e) {
       _logger.e('error starting session: $e');
+      reset();
```
jckras, line 68: is this where you start a new session upon failing? and what does failing mean in this context? same with starting a new session?

jckras (approved):
I have a few questions for my own understanding but other than that LGTM!

example/viam_example_app/lib/screens/stream.dart
clintpurser: because I replaced it with the widget which has this all in a reusable class

lib/src/robot/sessions_client.dart
clintpurser: I added a call to `metadata()` inside reset(), that is where it starts a new session upon failing.

njooma (approved):

PR #71, opened 2023-07-26
Prepare for publish

njooma:
Removed CHANGELOG.md because we don't keep that up to date

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -17,6 +17,7 @@ dependencies:
   logger: ^1.3.0
   fixnum: ^1.1.0
   flutter_joystick: ^0.0.3
+  collection: ^1.17.1
 
 dev_dependencies:
   flutter_test:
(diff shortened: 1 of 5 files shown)


pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   logger: ^1.3.0
   fixnum: ^1.1.0
   flutter_joystick: ^0.0.3
+  collection: ^1.17.1
```
clintpurser, line 20: I can't tell if this is being used in your changes, its not as far as I can tell, do you plan on using it soon?

njooma: It was already being used in WebRTC, but it was never depended on explicitly

clintpurser: ah i see. [here](https://github.com/viamrobotics/viam-flutter-sdk/blob/bde0240b5a44acfdf49843546281e6838970b70c/lib/src/rpc/web_rtc/web_rtc_transport_stream.dart#L3) 

sweet, thanks for cleaning up.

clintpurser (commented):
belated shipit

PR #72, opened 2023-07-26
Update README

njooma:


README.md
```diff
@@ -7,32 +8,21 @@ Build and connect to robots with Flutter
 ## (In)stability Notice
 
 > **Warning**
-> This is an alpha release of the Viam Flutter SDK. The only guarantee is that there **WILL** be breaking changes to the API. We will enumerate these changes in the release notes.
+> This is a beta release of the Viam Flutter SDK. No guarantees are made to the stability of the API. All changes will be enumerated in the release notes.
```
clintpurser, line 11: 🚨 🚨 🚨 BETA 🚨 🚨 🚨

jckras, line 11: What is the difference between alpha and beta release?

clintpurser (approved):

README.md
njooma: I've saying it's now `beta` since it's published to pub.dev rather than just on github. Otherwise, not functionally different.

clintpurser: [see here](https://en.wikipedia.org/wiki/Software_release_life_cycle#:~:text=Alpha%20testing%20is%20the%20first,the%20organization%20that%20developed%20it.)

Alpha is the first phase, typically smaller, lots of iteration. Beta is after, can be a larger group of people testing / slightly more public.

jckras (approved):
LGTM!  (q) question for my own understanding: Does this mean I need to update the Minimum iOS/android target in the mobile app or is this handled automatically by installing the viam-flutter-sdk? (also, If this is the wrong place to ask that question lmk)

njooma:
@jckras let's talk offline/slack

clintpurser:
> LGTM! (q) question for my own understanding: Does this mean I need to update the Minimum iOS/android target in the mobile app or is this handled automatically by installing the viam-flutter-sdk? (also, If this is the wrong place to ask that question lmk)

@jckras Apps manually track the minimum os versions.

README.md
jckras: Makes sense, thank you both :)

PR #73, opened 2023-07-27
Add workflow to update protos

njooma:
`sed` on macos works a little differently than `gnu sed`

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -1,0 +1,47 @@
+name: Update Protos + Open PR
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types:
+      - protos-updated
+
+jobs:
+  update-protos:
+    if: github.repository_owner == 'viamrobotics'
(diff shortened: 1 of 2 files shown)


maximpertsov (approved):

.github/workflows/update_protos.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Generate buf
+        run: make buf
+        env:
+          BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
+
+      - name: Format
+        run: make format
+
+      - name: Add + Commit + Open PR
+        uses: peter-evans/create-pull-request@v5
```
benjirewis, line 38: Still blows my mind that this is not an action provided by GitHub lol who is peter evans.

benjirewis (approved):

.github/workflows/update_protos.yml
njooma: Right?? I think you can manually do it since actions always include the GH command line tool but still

PR #78, opened 2023-07-31
[RSDK-4265] Widgets

njooma:
Adding widgets to the flutter sdk. Eventually this will be built out more, so I will be using this branch to add more widgets and merge them in.


lib/widgets/button.dart
```diff
@@ -0,0 +1,131 @@
...
+  ghost;
+}
+
+enum ViamButtonVariant {
+  iconOnly,
+  iconLeading,
+  iconTrailing;
+}
+
+class ViamButton extends StatelessWidget {
```
jckras, line 61: what is the use of a viamButton? does this mean rather than someone using an ElevatedButton from flutter and styling it to match the viam theme, then they could just use a viamButton? Or is there greater functionality?

example/viam_example_app/lib/screens/base.dart
```diff
@@ -1,15 +1,16 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
 import 'package:viam_sdk/viam_sdk.dart';
+import 'package:viam_sdk/widgets.dart';
 
 class BaseScreen extends StatelessWidget {
   final Base base;
   final ResourceName resourceName;
```
jckras, line 8: Do you need resourceName still? is it being used anywhere?

jckras (approved):
Added one potential change and the other comments are for my understanding, LGTM

lib/widgets/button.dart
njooma: It's basically just that -- just a button that's always going to be styled for Viam (regardless of what your theme is). I mostly made this for use internally for any/all buttons in the SDK widgets, and figured we might as well publish it like we do for PRIME.

example/viam_example_app/lib/screens/base.dart
njooma: We don't. I left it in because all the other screens still have `ResourceName`. I can remove it.

PR #79, opened 2023-08-01
RSDK-4058 Add PowerSensor Component

oliviamiller:
scope: https://docs.google.com/document/d/1HjShXMyD6HQkZkAo0bvDe61STdxxfyBrVnbRC6nSlM0/edit


lib/src/components/power_sensor/power_sensor.dart
```diff
@@ -0,0 +1,84 @@
...
+class Current {
+  double amperes;
+  bool isAc;
+
+  Current(this.amperes, this.isAc);
+
+  factory Current.fromProto(GetCurrentResponse pbResponse) {
+    return Current(pbResponse.amperes, pbResponse.isAc);
+  }
+}
```
njooma, line 27: Instead of creating these types which are identical to the response type (but with better names), you can just create `typedef`s that should do the same thing:

```dart
typedef Voltage = GetVoltageResponse;
typedef Current = GetCurrentResponse;
```

You will have to change a bit of the code, but this way you don't have to create brand new classes

njooma (changes requested):
Looks good! Just one change, and also let's get it up to date with `main`

njooma (approved):
Nice!

PR #80, opened 2023-08-02
[RSDK-4265] Board and Sensor Widgets

njooma:
New board widget and sensor widget.

Board widget updates the analogs/digital interrupts to use data tables rather than just showing the dict:
Screenshot 2023-08-02 at 1 49 46 PM

Sensor widget adds some new controls, along with auto refreshing:
Screenshot 2023-08-02 at 1 43 32 PM

Finally, updated the buttons so that they look nicer when it's icon only


example/viam_example_app/lib/screens/board.dart
```diff
@@ -1,94 +1,31 @@
...
-                ),
-                const Spacer()
-              ],
-            ),
-            const SizedBox(height: 16),
-            PlatformElevatedButton(
-              child: const Text('Set Pin State'),
-              onPressed: () => widget.board.setGpioState(setPin, high),
-            )
+            ViamBoardWidget(board: board)
```
jckras, line 28: awesome how much code this saves

lib/widgets/button.dart
```diff
@@ -1,12 +1,22 @@
...
+  /// Inverse of the primary button, Dark gray background, light gray foreground (inverse when in dark mode)
   inverse,
+
+  /// A button to indicate a successful operation, will result in a green color scheme
   success,
-  danger,
-  warning;
+
+  /// A button to indicate a warning state, will result in an amber color scheme
+  warning,
```
jckras, line 16: What is an example of something that is in a warning state? Just trying to understand the use case

njooma: Maybe an action that's not fully destructive, but just a warning. Maybe something like clearing logs in the future?

example/viam_example_app/lib/screens/base.dart
```diff
@@ -5,13 +5,11 @@ import 'package:viam_sdk/widgets.dart';
 
 class BaseScreen extends StatelessWidget {
   final Base base;
-  final ResourceName resourceName;
   final Iterable<Camera> cameras;
   final RobotClient robot;
 
   // TODO change BaseScreen to accept camera ResourceName.
```
benjirewis, line 11: Remove?

example/viam_example_app/lib/main.dart
```diff
@@ -92,8 +94,8 @@ class _MyHomePageState extends State<MyHomePage> {
       _loading = true;
     });
     final robotFut = RobotClient.atAddress(
-      '<URL>',
-      RobotClientOptions.withLocationSecret('<SECRET>'),
+      dotenv.env['ROBOT_LOCATION'] ?? '',
```
jckras, line 97: what would be stored in these env files? would a user upload their robot_location to the file or is this just a more professional way to include input that may change depending on user?

jckras (approved):
added a q for my own understanding but, LGTM

example/viam_example_app/lib/main.dart
njooma: Usually, environment-specific variables are stored in env files (in this case, the robot location and secret). This is super useful when you have different variables based on launch environment (e.g. debug vs release). This is useful in this example because it's also how we're testing internally, so it let's us set these values without the fear of committing them.

benjirewis (commented):
Any reason why `Link` and `Level` values in the images you included have values that end with `.`?

njooma:
> Any reason why `Link` and `Level` values in the images you included have values that end with `.`?

That's actually the value we're getting back from the server

benjirewis:
> > Any reason why `Link` and `Level` values in the images you included have values that end with `.`?
> 
> 
> 
> That's actually the value we're getting back from the server

Is that a server bug?

benjirewis (approved):

njooma:
> > > Any reason why `Link` and `Level` values in the images you included have values that end with `.`?
> > 
> > 
> > That's actually the value we're getting back from the server
> 
> Is that a server bug?

I actually don't know -- been trying to test that rover using the web control UI but it's breaking :(

PR #81, opened 2023-08-02
[RSDK-4265] Sensor widget tests

njooma:
Documentation and tests for sensor widget

### DIFF
--- a/test/widget_tests/sensor_widget_test.dart
+++ b/test/widget_tests/sensor_widget_test.dart
@@ -80,9 +80,11 @@ void main() {
 
       final playButton = find.widgetWithIcon(ViamButton, Icons.play_arrow);
       final pauseButton = find.widgetWithIcon(ViamButton, Icons.pause);
+      final refreshButton = find.widgetWithIcon(ViamButton, Icons.refresh);
 
       expect(playButton, findsNothing);
       expect(pauseButton, findsNothing);
+      expect(refreshButton, findsOneWidget);
     });
   });
 }
--- a/test/widget_tests/sensor_widget_test.dart
+++ b/test/widget_tests/sensor_widget_test.dart
@@ -1,0 +1,77 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:viam_sdk/widgets.dart';
+
+import '../test_utils.dart';
+import '../unit_test/components/sensor_test.dart';
+
+void main() {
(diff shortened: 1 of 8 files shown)


test/widget_tests/sensor_widget_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      await tester.pumpWidget(widget);
+
+      final playButton = find.widgetWithIcon(ViamButton, Icons.play_arrow);
+      final pauseButton = find.widgetWithIcon(ViamButton, Icons.pause);
+      final refreshButton = find.widgetWithIcon(ViamButton, Icons.refresh);
+
+      expect(playButton, findsNothing);
+      expect(pauseButton, findsNothing);
+      expect(refreshButton, findsNothing);
+    });
```
jckras, line 75: how come you aren't calling tester.pump() in this test like you are in the other ones?

test/widget_tests/sensor_widget_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      final widget = TestableWidget(child: ViamSensorWidget(sensor: FakeSensor('sensor'), refreshInterval: null));
+      await tester.pumpWidget(widget);
+
+      final playButton = find.widgetWithIcon(ViamButton, Icons.play_arrow);
+      final pauseButton = find.widgetWithIcon(ViamButton, Icons.pause);
+
+      expect(playButton, findsNothing);
+      expect(pauseButton, findsNothing);
+    });
+  });
```
jckras, line 87: (q) maybe obvious but does it also hide the refresh button? should we still test for that?

jckras (approved):
two questions about the test but LGTM

test/widget_tests/sensor_widget_test.dart
njooma: `tester.pump` only needs to be called if the state changes in order to reevaluate the UI with the new state [see docs](https://docs.flutter.dev/cookbook/testing/widget/introduction#notes-about-the-pump-methods)

test/widget_tests/sensor_widget_test.dart
njooma: Nope! When `refreshInterval` is `null`, only the play/pause button is hidden. The manual refresh button **needs** to be available so that users can manually refresh

njooma: But you are correct that we should have a test to make sure the refresh button **is** visible so I added it :)

PR #96, opened 2023-08-07
Reconnect camera streams

njooma:
Camera streams now automatically reestablish after a loss and regain of connection.

### DIFF
--- a/lib/src/media/stream/client.dart
+++ b/lib/src/media/stream/client.dart
@@ -1,15 +1,17 @@
 import 'dart:async';
 
 import 'package:flutter_webrtc/flutter_webrtc.dart';
+import 'package:logger/logger.dart';
 
 import '../../gen/proto/stream/v1/stream.pbgrpc.dart';
 import '../../rpc/web_rtc/web_rtc_client.dart';
+
+Logger _logger = Logger();
 
 class StreamManager {
   final Map<String, MediaStream> _streams = {};
   final Map<String, StreamClient> _clients = {};
   WebRtcClientChannel _channel;
-  final StreamServiceClient _client;
   // ignore: cancel_subscriptions
   StreamSubscription? _errorHandler;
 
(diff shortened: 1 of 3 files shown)


cheukt (approved):

lib/src/media/stream/client.dart
```diff
@@ -1,36 +1,45 @@
...
 
+  StreamServiceClient get _client {
+    return StreamServiceClient(_channel);
+  }
+
   set channel(WebRtcClientChannel channel) {
+    _errorHandler?.cancel();
     _channel = channel;
     _channel.rtcPeerConnection.onTrack = (event) {
+      _errorHandler?.cancel(); // Cancel the error handler -- clearly we're connected if we're receiving this event
```
jckras, line 35: (q) why are we doing this twice?

lib/src/robot/client.dart
```diff
@@ -75,7 +80,10 @@ class RobotClient {
   /// Refresh the resources of this robot
   Future<void> refresh() async {
     final ResourceNamesResponse response = await _client.resourceNames(ResourceNamesRequest());
-    if (response.resources == resourceNames) {
+    if (setEquals(response.resources.toSet(), resourceNames.toSet())) {
```
jckras, line 83: (q) why `if(setEquals(response.resources.toSet(),.......` as opposed to `if (response.resources == resourceNames)` ?

lib/src/media/stream/client.dart
njooma: The first time happens whenever we change the channel.

This one is on a callback -- so it's possible it gets called out of order/after the error handler exists again.

njooma: The first one happens whenever the channel changes. This one is happening on a callback which could be called out of order/delayed (after  the `_errorHandler` exists again). The one that happens a bit further down is in another callback, where we want to cancel if the returned state is not an error state.

lib/src/robot/client.dart
njooma: Mostly because the second way wasn't working. I didn't dive too deep into it but converting both to lists, sorting those lists, making everything lowercase strings, nothing actually make things equal except converting these to sets and using `setEquals`

PR #100, opened 2023-08-16
RSDK-4580: Flutter SDK Base Properties

martha-johnston:
Update Flutter SDK to include base properties


martha-johnston:
I think same thing is happening here as with python... checks aren't passing because it's not using the most recent api. is that something I need to update in this PR?

lib/src/components/base/base.dart
```diff
@@ -1,7 +1,20 @@
...
+  double widthMeters;
+  double wheelCircumferenceMeters;
+
+  BaseProperties(this.turningRadiusMeters, this.widthMeters, this.wheelCircumferenceMeters);
+
+  factory BaseProperties.fromProto(GetPropertiesResponse pbResponse) {
+    return BaseProperties(pbResponse.turningRadiusMeters, pbResponse.widthMeters, pbResponse.wheelCircumferenceMeters);
+  }
+}
+
```
njooma, line 17: Instead of creating a class that mirrors the existing type, you can use a `typedef` similar to [here](https://github.com/viamrobotics/viam-flutter-sdk/blob/82dbe48c1a26e28fe6ab6e195a1e6dea1bb7d9b1/lib/src/components/power_sensor/power_sensor.dart#L7)

njooma (changes requested):
I've merged in the new protobufs, so you should rebase main to get the new protos

lib/src/components/base/base.dart
martha-johnston: Yeah will do, but wondering why you want this Properties type to be `typedef`? There are only two other properties (motor and camera) but they are both set up as classes

njooma: Great question -- likely because we did those earlier without realizing we didn't need to create custom classes for it. Those should actually be swapped to `typdef` as well.

IMO we should use `typedef` because they will automatically stay in sync with the protos should anything changed/added/removed

martha-johnston: do you want me to change the other two while I'm at it or would you rather they be done separately?

njooma: I mean, if you want to do it that'll save some time and an additional PR! Thanks :)

PR #101, opened 2023-08-16
Fix where reconnection might've gotten stuck in a loop

njooma:
It's possible that the previous way for reconnecting would attempt to reconnect multiple times while other reconnection attempts were already in flight.

I've changed the functionality to use a boolean check instead of canceling timers in order to break out of potential cycles.

### DIFF
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -178,10 +178,11 @@ class RobotClient {
       } catch (e) {
         await _channel.shutdown();
         _sessionsClient.reset();
-        _logger.i('Failed to reconnect, retrying in $reconnectInterval second${reconnectInterval != 1 ? "s" : ""}');
         if (!_shouldAttemptReconnection) {
+          _logger.i('Failed to reconnect. No more attempts to reconnect will be made.');
           break;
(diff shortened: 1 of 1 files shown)


lib/src/robot/client.dart
```diff
@@ -153,33 +154,36 @@ class RobotClient {
...
+        _client = client;
+        _sessionsClient = SessionsClient(_channel, _options.enableSessions);
+        await refresh();
+        _connected = true;
+        _logger.i('Successfully reconnected to robot');
+        _startCheckConnectionTask();
+      } catch (e) {
+        await _channel.shutdown();
+        _sessionsClient.reset();
+        _logger.i('Failed to reconnect, retrying in $reconnectInterval second${reconnectInterval != 1 ? "s" : ""}');
```
stuqdog, line 181: (minor) seems a little strange that we would log this even if `_shouldAttemptReconnection` is false, can we move this log message below the `break` and create a separate log when we aren't attempting anymore?

stuqdog (approved):
One small thing (no need to rerequest review) otherwise lgtm!

PR #109, opened 2023-08-29
[RSDK-4323] Gripper

njooma:
Add gripper as an available component, and also add its widget.

Also:
* Extract the camera stream and picker widget into its own component so it can be reused


njooma:
Looks like:
IMG_0E133CE16745-1

cheukt (approved):

PR #111, opened 2023-09-11
APP-2521: SDK: Update Flutter code sample authing with API key

ohEmily:
[APP-2521](https://viam.atlassian.net/browse/APP-2521)

TODO:
 - check if there are other callsites to update

[APP-2521]: https://viam.atlassian.net/browse/APP-2521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -39,7 +39,7 @@ class RobotClientOptions {
   RobotClientOptions.withDialOptions(this.dialOptions);
 
   // Convenience initializer for creating options with an API key ID/API key pair
-  RobotClientOptions.withAPIKey(String apiKeyID, String apiKey)
+  RobotClientOptions.withApiKey(String apiKeyID, String apiKey)
       : dialOptions = DialOptions()
           ..credentials = Credentials.apiKey(apiKey)
           ..authEntity = apiKeyID;
--- a/lib/src/robot/client.dart
+++ b/lib/src/robot/client.dart
@@ -38,6 +38,12 @@ class RobotClientOptions {
   /// Convenience initializer for creating options with specified [DialOptions]
   RobotClientOptions.withDialOptions(this.dialOptions);
 
+  // Convenience initializer for creating options with an API key ID/API key pair
+  RobotClientOptions.withAPIKey(String apiKeyID, String apiKey)
+      : dialOptions = DialOptions()
+          ..credentials = Credentials.apiKey(apiKey)
+          ..authEntity = apiKeyID;
+
   /// Convenience initializer for creating options with a robot location secret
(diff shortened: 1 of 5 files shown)


lib/src/robot/client.dart
```diff
@@ -38,6 +38,12 @@ class RobotClientOptions {
   /// Convenience initializer for creating options with specified [DialOptions]
   RobotClientOptions.withDialOptions(this.dialOptions);
 
+  // Convenience initializer for creating options with an API key ID/API key pair
+  RobotClientOptions.withAPIKey(String apiKeyID, String apiKey)
```
njooma, line 42: this should be `withApiKey` ([Dart style guide](https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words) for initialisms longer than 2 letters)

njooma (approved):
Looks good! One small comment, but nice work overall

lib/src/robot/client.dart
ohEmily: oh thank you so much for catching this!

ohEmily:
hey @njooma, just found `/Users/emily/viam/viam-flutter-sdk/lib/src/viam_sdk_impl.dart` and `/Users/emily/viam/viam-flutter-sdk/lib/src/viam_sdk.dart`. Is this what external users are supposed to use? How widespread is usage? I think it's OK for me to leave `withAccessToken` as-is (this is the client JWT returned, right?), but I'm tempted to update `getRobotClient` to use an API key and only fallback to robot secrets if API keys are not available.

njooma:
This is what apps use to access viam using JWTs from auth0. I don't think anything needs to change here, including `getRobotClient` since the user isn't providing the secret, but the SDK is getting it behind the scenes using the APIs. 

Is there a way to programmatically get a robot's API key?

benjirewis (approved):
LGTM but I don't have much context here.

ohEmily:
> This is what apps use to access viam using JWTs from auth0. I don't think anything needs to change here, including getRobotClient since the user isn't providing the secret, but the SDK is getting it behind the scenes using the APIs.
>
> Is there a way to programmatically get a robot's API key?

Hmm we definitely want to move towards getting SDKs off of authenticating with secrets as they don't respect our permissions model, so I'd like to move towards that in the future. But the API to list keys isn't currently merged, so I'm going to merge this for now and file a follow-up ticket for having SDKs authenticate with API keys (keeping the fallback to secrets for now).

PR #125, opened 2023-10-04
Remove delayed stop check

njooma:
10ms is faster than any human reaction time, so it's virtually impossible that the user let go of the slider, then reengaged it within the 10ms time frame. I think it's safe to remove that check and just stop the motor again if it's still moving.

I'm not sure the underlying issue here, but using print statements, basically I found that the `setState(() { power = 0; });` sometimes wasn't actually setting the power to zero. Not entirely sure what's going on there, but I'm not going to fight the system. 

Also, removed the `setState` call in `_handleSliderRelease`, which is called `onChangeEnd`, because the [docs](https://api.flutter.dev/flutter/material/Slider/onChangeEnd.html) say:
> This callback shouldn't be used to update the slider [value](https://api.flutter.dev/flutter/material/Slider/value.html) (use [onChanged](https://api.flutter.dev/flutter/material/Slider/onChanged.html) for that)

### DIFF
--- a/lib/widgets/resources/motor.dart
+++ b/lib/widgets/resources/motor.dart
@@ -42,10 +42,10 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
       });
       await Future.delayed(const Duration(milliseconds: 10));
       if (await widget.motor.isMoving()) {
-        if (power == 0) {
-          await widget.motor.stop();
-          setState(() {});
-        }
+        await widget.motor.stop();
+        setState(() {
+          power = 0;
+        });
       }
     } catch (e) {
       error = e as Error;
@@ -53,9 +53,6 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
   }
 
   void _handleSliderRelease(double power) {
-    setState(() {
-      this.power = power;
-    });
     if (autoStop) {
       stop();
     }


lib/widgets/resources/motor.dart
```diff
@@ -42,20 +42,17 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
       });
       await Future.delayed(const Duration(milliseconds: 10));
       if (await widget.motor.isMoving()) {
-        if (power == 0) {
```
benjirewis, line 45: Trying to understand what's going on here. So the previous behavior of `stop` was that we'd try stopping the motor (and setting power state on the widget to 0). Then, we'd wait 10ms. Then, we'd see if the motor was still moving, and only attempt a `stop` again if the power state was still 0? Assuming the user could have altered the power state within those 10ms and no longer want the motor to stop?

Why do we need to wait 10ms at all? Why can't the definition of this method just be:
```
Future<void> stop() async {
  try {
    await widget.motor.stop();
    setState(() {
      power = 0;
    });
  } catch (e) {
    error = e as Error;
  }
}
```

clintpurser: If I remember correctly the the 10ms pause and then checking again was an attempt to solve this same bug when it was present in the example app. Back then it wasn't honoring the stop call sometimes when the user released their touch, so this was an attempt to ensure it actually stopped.

benjirewis: Gotcha. I guess I wonder why the first stop doesn't cut it, but the workaround makes sense.

lib/widgets/resources/motor.dart
```diff
@@ -42,20 +42,17 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
       });
       await Future.delayed(const Duration(milliseconds: 10));
```
clintpurser, line 43: This was my fault, I should have left a comment originally explaining why we had this odd delay and a re attempt at stopping. Perhaps we include something like this?

```suggestion
      });
      // Sometimes the motor does not honor the first Stop call
      // So we wait a small amount of time and try again.
      await Future.delayed(const Duration(milliseconds: 10));
```

lib/widgets/resources/motor.dart
```diff
@@ -42,20 +42,17 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
...
       }
     } catch (e) {
       error = e as Error;
     }
   }
 
   void _handleSliderRelease(double power) {
-    setState(() {
-      this.power = power;
-    });
```
clintpurser, line 58: 👍  good catch on the widgets documentation

benjirewis (approved):
LGTM; I have like no context here. @njooma's comment is def concerning: 

> basically I found that the setState(() { power = 0; }); sometimes wasn't actually setting the power to zero.

but I certainly won't block the PR on understanding why that was happening.

clintpurser (approved):
LGTM, good catch, thank you!

I pulled your branch into the app and tested before/after and this does appear to resolve the issue.

Before:

https://github.com/viamrobotics/viam-flutter-sdk/assets/22106496/c8f45ac7-efba-4c42-9a2e-1147b9089c75

After:

https://github.com/viamrobotics/viam-flutter-sdk/assets/22106496/2b2787ee-5b3f-4629-b6ef-b6263ae839f2

One thought on the app side, we should lock down scrolling while pointer is down for these sliders too, just like the base widget.

lib/widgets/resources/motor.dart
clintpurser: My theory is because of the large amount of setPower calls we're making. maybe somehow that overloads things and makes a race condition. We could limit those calls and look for a certain percentage of change or allow so many calls per second if thats the case, which might be a better solution. But I haven't looked into it enough to confirm.

clintpurser:
yeah interesting about  setState(() { power = 0; }); - i wonder if that needs to be called before stop(), otherwise the old value persists long enough to muddle the calls?

cheukt (approved):

lib/widgets/resources/motor.dart
njooma: I wish I knew. My tests show that the system just doesn't set the power to zero, or that there's some weird touch behavior where the power is set to zero, but the UI pushes it back to whatever it used to be and negates the release.

PR #131, opened 2023-10-13
Arm widget

njooma:
New arm widget.

Also, update buttons so we can actually have small sizes.

simulator_screenshot_F2989544-18A4-4943-BB5D-B703BF748910

### DIFF
--- a/lib/src/components/arm/arm.dart
+++ b/lib/src/components/arm/arm.dart
@@ -1,5 +1,4 @@
 import '../../gen/common/v1/common.pb.dart';
-import '../../gen/component/arm/v1/arm.pb.dart';
 import '../../resource/base.dart';
 import '../../robot/client.dart';
 
@@ -13,11 +12,11 @@ abstract class Arm extends Resource {
   /// Move the end of the arm to the [Pose] specified.
   Future<void> moveToPosition(Pose pose, {Map<String, dynamic>? extra});
 
-  /// Move each joint on the arm to the corresponding position specified in [JointPositions]
-  Future<void> moveToJointPositions(JointPositions positions, {Map<String, dynamic>? extra});
+  /// Move each joint on the arm to the corresponding position specified in [positions]
+  Future<void> moveToJointPositions(List<double> positions, {Map<String, dynamic>? extra});
 
-  /// Get the [JointPositions] representing the current position of the arm
-  Future<JointPositions> jointPositions({Map<String, dynamic>? extra});
+  /// Get the [List] representing the current position of the arm
+  Future<List<double>> jointPositions({Map<String, dynamic>? extra});
 
   /// Stops all motion of the arm. It is assumed that the arm stops immediately.
   Future<void> stop({Map<String, dynamic>? extra});
--- a/lib/widgets/button.dart
+++ b/lib/widgets/button.dart
@@ -85,6 +85,7 @@ enum ViamButtonFillStyle {
 
 /// The size class of the button.
 enum ViamButtonSizeClass {
+  xxs,
   xs,
   small,
   medium,
@@ -94,6 +95,8 @@ enum ViamButtonSizeClass {
   /// The font size of the button, based on size class
   double get fontSize {
     switch (this) {
+      case xxs:
+        return 10;
       case xs:
         return 10;
       case small:
@@ -110,6 +113,8 @@ enum ViamButtonSizeClass {
   /// The padding of the button, based on size class
   EdgeInsets get padding {
     switch (this) {
+      case xxs:
+        return const EdgeInsets.all(0);
       case xs:
         return const EdgeInsets.symmetric(vertical: 4, horizontal: 8);
       case small:
@@ -125,7 +130,12 @@ enum ViamButtonSizeClass {
 
   /// The style of the button, based on size class
   ButtonStyle get style {
-    return ButtonStyle(textStyle: MaterialStatePropertyAll(TextStyle(fontSize: fontSize)), padding: MaterialStatePropertyAll(padding));
+    return ButtonStyle(
+      textStyle: MaterialStatePropertyAll(TextStyle(fontSize: fontSize)),
+      padding: MaterialStatePropertyAll(padding),
+      tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+      minimumSize: const MaterialStatePropertyAll(Size.zero),
+    );
   }
 }
 
--- a/lib/widgets/resources/arm.dart
+++ b/lib/widgets/resources/arm.dart
@@ -1,0 +1,281 @@
+import 'package:collection/collection.dart';
+import 'package:flutter/material.dart';
+import 'package:viam_sdk/protos/component/arm.dart';
+import 'package:viam_sdk/viam_sdk.dart';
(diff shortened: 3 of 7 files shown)


njooma:
I also changed the return type of `Arm.jointPositions` from the proto `JointPositions` to a `List<double>`. I feel this is more in line with our system of trying to hide the proto definitions in favor of native types where possible, especially since the proto `JointPositions` had only one attribute: a `values` property which is a `List<double>`.

lib/src/components/arm/arm.dart
```diff
@@ -13,11 +12,11 @@ abstract class Arm extends Resource {
   /// Move the end of the arm to the [Pose] specified.
   Future<void> moveToPosition(Pose pose, {Map<String, dynamic>? extra});
 
-  /// Move each joint on the arm to the corresponding position specified in [JointPositions]
-  Future<void> moveToJointPositions(JointPositions positions, {Map<String, dynamic>? extra});
+  /// Move each joint on the arm to the corresponding position specified in [positions]
+  Future<void> moveToJointPositions(List<double> positions, {Map<String, dynamic>? extra});
```
benjirewis, line 16: Nice, good catch here; definitely agree with preferring native types where possible.

lib/widgets/button.dart
```diff
@@ -94,6 +95,8 @@ enum ViamButtonSizeClass {
   /// The font size of the button, based on size class
   double get fontSize {
     switch (this) {
+      case xxs:
+        return 10;
```
benjirewis, line 99: [q] why the same font size as `xs`?

lib/widgets/resources/base.dart
```diff
@@ -34,8 +34,8 @@ class _ViamBaseWidgetState extends State<ViamBaseWidget> {
 
   @override
   void initState() {
-    camera = widget.cameras.firstOrNull;
     super.initState();
+    camera = widget.cameras.firstOrNull;
```
benjirewis, line 38: [q] why move this after the super init state? Was it getting overwritten or something?

benjirewis (approved):
I don't have too much context here... just some qs, but looks p good from what I can tell.

lib/widgets/button.dart
njooma: Anything smaller than 10 becomes pretty difficult to read

lib/widgets/resources/base.dart
njooma: We've just been doing it wrong the entire time. From the [documentation](https://api.flutter.dev/flutter/widgets/State/initState.html):

> Implementations of this method should start with a call to the inherited method, as in super.initState().

njooma: Similarly, you might see changes to `super.dispose()` in other PRs (not in this one) for the inverse reason:

> Implementations of this method should end with a call to the inherited method, as in super.dispose().

benjirewis: Cool; makes intuitive sense too that you want to do the inherited construction/destruction before the custom class construction/destruction.

lib/widgets/resources/arm.dart
```diff
@@ -0,0 +1,280 @@
...
+
+  @override
+  State<ViamArmWidget> createState() => _ViamArmWidgetState();
+}
+
+class _ViamArmWidgetState extends State<ViamArmWidget> {
+  Pose endPosition = Pose();
+  List<double> jointPositions = [];
+
+  Future<void> _positions() async {
```
clintpurser, line 24: [minor] I think in this case it is okay to prefix the method name with `get`.  omitting `get` / `set` is more for API design, since this the internal workings of a widget and isn't a public API I think naming the method more clearly would be helpful.

```suggestion
  Future<void> _getPositions() async {
```

ref: https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

lib/widgets/resources/arm.dart
```diff
@@ -0,0 +1,280 @@
...
+    } else if (field == 'ox') {
+      ep.oX += increment;
+    } else if (field == 'oy') {
+      ep.oY += increment;
+    } else if (field == 'oz') {
+      ep.oZ += increment;
+    }
+    await widget.arm.moveToPosition(ep);
+    await _positions();
+  }
```
clintpurser, line 58: [suggestion] I think the string comparison here is _fine_, I might suggest defining an enum for the `fields` with the different options, and using a switch statement instead.

Not something blocking merging, just a preference on my side.

lib/widgets/resources/arm.dart
```diff
@@ -0,0 +1,280 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.center,
+      crossAxisAlignment: CrossAxisAlignment.center,
+      children: [
+        const Text('End Positions (mm)', style: TextStyle(fontWeight: FontWeight.bold)),
+        Table(
```
clintpurser, line 74: [minor] This Table is a BEEFY widget, can we break it out into its own widget?

clintpurser (changes requested):
I think this needs some more work, close though!

lib/widgets/resources/arm.dart
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  Future<void> updateJointPosition(int joint, double increment) async {
+    final jp = jointPositions;
+    jp[joint] += increment;
+    await widget.arm.moveToJointPositions(jp);
+    await _getPositions();
+  }
+
+  TableRow _getEndPositionRow(_PoseField field) {
```
clintpurser, line 98: Nice! Can we put this in it's own widget instead of a helper function? See [here](https://steveos.medium.com/use-separate-widget-over-helper-method-in-flutter-heres-why-better-performace-433672eb7461) for a bit more of a deep dive on why. Basically it's more performant when the widget tree rebuilds.

lib/widgets/resources/arm.dart
```diff
@@ -0,0 +1,192 @@
...
+          defaultVerticalAlignment: TableCellVerticalAlignment.middle,
+          children: _PoseField.values.map((e) => _getEndPositionRow(e)).toList(),
+        ),
+        const SizedBox(height: 16),
+        const Text('Joints (degrees)', style: TextStyle(fontWeight: FontWeight.bold)),
+        Table(
+          columnWidths: const {0: IntrinsicColumnWidth()},
+          defaultVerticalAlignment: TableCellVerticalAlignment.middle,
+          children: jointPositions
+              .mapIndexed((index, element) => TableRow(children: [
```
clintpurser, line 159: I think this `TableRow` should still be broken out into it's own widget as well.

clintpurser (changes requested):
similar feedback, overall closer, but need to prefer widgets over helper functions.

lib/widgets/resources/arm.dart
clintpurser: @njooma and I spoke offline, I was mistaken and TableRow is not a `Widget` (<--- WHAT?!) so I was mistaken. this helper function in this case is fine :)

clintpurser (approved):
LGTM!

PR #133, opened 2023-10-18
Add permissions APIs and values

njooma:
We need permissions in the SDK for proactive checks around auth

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import '../gen/app/v1/app.pbgrpc.dart';
+import 'permissions.dart';
 
 typedef RobotPartLogPage = GetRobotPartLogsResponse;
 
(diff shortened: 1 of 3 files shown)


lib/src/app/app.dart
```diff
@@ -86,4 +87,25 @@ class AppClient {
     final stream = response.map((event) => event.logs);
     return stream.asBroadcastStream(onCancel: (_) => response.cancel());
   }
+
+  /// List the [Authorization]s available for the currently authenticated user
+  Future<List<Authorization>> listAuthorizations(String organizationId, {List<String> resourceIds = const []}) async {
+    final request = ListAuthorizationsRequest()
+      ..organizationId = organizationId
+      ..resourceIds.addAll(resourceIds);
```
maximpertsov, line 95: [`..`](https://dart.dev/language/operators#cascade-notation) is neat a language feature

lib/src/app/app.dart
```diff
@@ -86,4 +87,25 @@ class AppClient {
...
+  Future<List<Permission>> checkPermissions(ResourceType resourceType, String resourceId, List<Permission> permissions) async {
+    final request = CheckPermissionsRequest()
+      ..permissions.add((AuthorizedPermissions()
+        ..resourceType = resourceType.name
+        ..resourceId = resourceId
+        ..permissions.addAll(permissions.map((e) => e.value))));
+    final response = await _client.checkPermissions(request);
+    return response.authorizedPermissions.first.permissions
+        .map((e) => Permission.values.firstWhere((element) => element.value == e))
+        .toList();
```
maximpertsov, line 109: [q] (mainly because I'm unfamiliar with the shape of the API response) - why are we only considering the first authorized permission in the response? (`response.authorizedPermissions.first`)

maximpertsov (commented):
have a question about the check api

lib/src/app/app.dart
njooma: This is a great question! It's because the API actually takes in an array of of permissions to check, and then returns an array of validated permissions. 

This method only allows for a single permission check, so we will only get (at most) a single permission response back

lib/src/app/app.dart
njooma: Ya SUPER useful in not having to create temporary objects for assignment

clintpurser (approved):
LGTM

PR #137, opened 2023-10-23
RSDK-5141 - Add MDNS support

clintpurser:
Add MDNS support to the Flutter SDK. 

I had to add our own version of the GrpcOrGrpcWebChannel wrappers to continue to allow web support.

### DIFF
--- a/lib/src/rpc/dial.dart
+++ b/lib/src/rpc/dial.dart
@@ -196,7 +196,8 @@ Future<ClientChannelBase> _dialWebRtc(String address, DialOptions options, Strin
   _logger.d('Connecting to signaling server: $signalingServer');
   final signalingChannel = await _dialDirectGrpc(signalingServer, options, sessionCallback);
   _logger.d('Connected to signaling server: $signalingServer');
-  final signalingClient = SignalingServiceClient(signalingChannel, options: CallOptions(metadata: {'rpc-host': options.authEntity!}));
+  final signalingClient = SignalingServiceClient(signalingChannel,
+      options: CallOptions(metadata: {'rpc-host': ((options.usingMdns) ? options.authEntity : address)!}));
   WebRTCConfig config;
   try {
     config = (await signalingClient.optionalWebRTCConfig(OptionalWebRTCConfigRequest())).config;
--- a/lib/src/rpc/dial.dart
+++ b/lib/src/rpc/dial.dart
@@ -47,8 +47,11 @@ class DialOptions {
   /// [externalAuthAddress], [externalAuthToEntity]
   String? accessToken;
 
-  /// Whether the connection was made usuing MDNS
-  bool usingMdns = false;
+  /// Wether an mDNS connection will be attempted
+  bool attemptMdns = true;
+
+  /// Whether the connection was made using mDNS
+  bool _usingMdns = false;
 }
 
 /// The credentials used for connecting to the robot
@@ -104,26 +107,25 @@ class DialWebRtcOptions {
   String? signalingAccessToken;
 }
 
-String allowedAddress = '';
-
 /// Connect to a robot at the provided address with the given options
 Future<ClientChannelBase> dial(String address, DialOptions? options, String Function() sessionCallback) async {
   _logger.i('Connecting to address $address');
   final opts = options ?? DialOptions();
-  String? mdnsUri;
 
-  try {
-    mdnsUri = await searchMdns(address);
-    // Let downstream calls know when mdns was used. This is helpful to inform
-    // when determining if we want to use the external auth credentials for the signaling
-    // in cases where the external signaling is the same as the external auth. For mdns
-    // this isn't the case.
-    final dialOptsCopy = opts
-      ..usingMdns = true
-      ..authEntity = address;
-    return _dialWebRtc(mdnsUri, dialOptsCopy, sessionCallback);
-  } catch (e) {
-    _logger.w('error dialing with mDNS; falling back to other methods', error: e);
+  if (opts.attemptMdns) {
+    try {
+      final mdnsUri = await searchMdns(address);
+      // Let downstream calls know when mdns was used. This is helpful to inform
+      // when determining if we want to use the external auth credentials for the signaling
+      // in cases where the external signaling is the same as the external auth. For mdns
+      // this isn't the case.
+      final dialOptsCopy = opts
+        .._usingMdns = true
+        ..authEntity = address;
+      return _dialWebRtc(mdnsUri, dialOptsCopy, sessionCallback);
+    } catch (e) {
+      _logger.d('Error dialing with mDNS; falling back to other methods', error: e);
+    }
   }
 
   bool disableWebRtc = opts.webRtcOptions?.disable ?? false;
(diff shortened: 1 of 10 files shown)


lib/src/rpc/grpc/grpc_or_grpcweb_channel_grpc.dart
```diff
@@ -0,0 +1,29 @@
...
+          port: grpcPort,
+          options: ChannelOptions(
+            credentials: grpcTransportSecure
+                ? ChannelCredentials.secure(onBadCertificate: (_, host) {
+                    // When making an MDNS connection the certificate will be rejected because
+                    // the local IP address doesn't match the certificate. Upon failure we
+                    // need to check if the host matches the IP we are attempting to connect to
+                    // and allow if it does.
+                    if (host == '$grpcHost:$grpcPort') return true;
+                    return false;
```
clintpurser, line 24: This onBadCertificate callback is the sole reason we need our own versions of the GrpcOrGrpcWebChannel classes. It is all set as final in the constructors and I couldn't figure out how else to inject it. Without this the certificates are rejected because the host doesn't match the certificate.

lib/src/rpc/dial.dart
```diff
@@ -139,11 +192,11 @@ Future<ClientChannelBase> _dialWebRtc(String address, DialOptions options, Strin
     }
   }
 
-  final signalingServer = options.webRtcOptions?.signalingServerAddress ?? 'app.viam.com';
+  final signalingServer = options.webRtcOptions?.signalingServerAddress ?? ((options.usingMdns) ? address : 'app.viam.com');
   _logger.d('Connecting to signaling server: $signalingServer');
   final signalingChannel = await _dialDirectGrpc(signalingServer, options, sessionCallback);
   _logger.d('Connected to signaling server: $signalingServer');
-  final signalingClient = SignalingServiceClient(signalingChannel, options: CallOptions(metadata: {'rpc-host': address}));
+  final signalingClient = SignalingServiceClient(signalingChannel, options: CallOptions(metadata: {'rpc-host': options.authEntity!}));
```
clintpurser, line 199: oop, I need to add logic to change this... incoming...

pubspec.yaml
```diff
@@ -19,6 +19,7 @@ dependencies:
   flutter_joystick: ^0.0.3
   collection: ^1.17.1
   intl: ^0.18.1
+  bonsoir: ^2.2.0+1 # do not upgrade beyond 2.2.0+1 until this issue is resolved: https://github.com/Skyost/Bonsoir/issues/57
```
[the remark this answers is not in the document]
njooma: Should we pin to exactly this version instead of using `^`?

lib/src/rpc/dial.dart
```diff
@@ -99,10 +104,28 @@ class DialWebRtcOptions {
   String? signalingAccessToken;
 }
 
+String allowedAddress = '';
```
njooma, line 107: What is `allowedAddress` for, and why is it global?

lib/src/rpc/dial.dart
```diff
@@ -44,6 +46,9 @@ class DialOptions {
   /// If enabled, other auth options have no affect. Eg. [authEntity], [credentials],
   /// [externalAuthAddress], [externalAuthToEntity]
   String? accessToken;
+
+  /// Whether the connection was made usuing MDNS
+  bool usingMdns = false;
```
njooma, line 51: Can this be private? IMO, `DialOptions` should be used to allow users to set options, not for internal state-keeping.

Also, I think we should have an option called `useMdns` that allows users to turn on/off mdns.

lib/src/rpc/dial.dart
```diff
@@ -99,10 +104,28 @@ class DialWebRtcOptions {
...
+    // Let downstream calls know when mdns was used. This is helpful to inform
+    // when determining if we want to use the external auth credentials for the signaling
+    // in cases where the external signaling is the same as the external auth. For mdns
+    // this isn't the case.
+    final dialOptsCopy = opts
+      ..usingMdns = true
+      ..authEntity = address;
+    return _dialWebRtc(mdnsUri, dialOptsCopy, sessionCallback);
+  } catch (e) {
+    _logger.w('error dialing with mDNS; falling back to other methods', error: e);
```
njooma, line 126: I think this should just be `info` or `debug` level, since the user can't do anything about it, it's just to let them know what happened.

njooma: (Also can we have it be `Error* dialing` for proper capitalization :) )

lib/src/rpc/dial.dart
```diff
@@ -128,6 +175,7 @@ Future<ClientChannelBase> _dialDirectGrpc(String address, DialOptions options, S
 }
 
 Future<ClientChannelBase> _dialWebRtc(String address, DialOptions options, String Function() sessionCallback) async {
+  allowedAddress = address;
```
njooma, line 178: Again from above, what is `allowedAddress` for?

lib/src/rpc/grpc/grpc_or_grpcweb_channel_grpc.dart
njooma: Damn, good find!

njooma (changes requested):

lib/src/rpc/dial.dart
```diff
@@ -113,6 +136,30 @@ Future<ClientChannelBase> dial(String address, DialOptions? options, String Func
   return _dialWebRtc(address, opts, sessionCallback);
 }
 
+Future<String> searchMdns(String address) async {
+  // We need to replace all periods with dashes, because this is how viam instances are broadcast locally.
```
stuqdog, line 140: (q) Is this always the case? I often see both the `.` and `-` versions when querying mDNS.

stuqdog (approved):
A couple minor spelling things and a question, otherwise lgtm (modulo Naveed's comments)!

pubspec.yaml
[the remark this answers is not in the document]
clintpurser: i _think_ we're fine using the latest version of 2.2.0 available - this won't bump to the next major version -- https://dart.dev/tools/pub/versioning#semantic-versions

lib/src/rpc/dial.dart
clintpurser: left over from a previous attempted implementation. Removing :)

lib/src/rpc/dial.dart
clintpurser: I have seen both as well, but with the bonsoir package I am using it was only finding the `-` version.

lib/src/rpc/dial.dart
clintpurser: leftover code i meant to remove. sorry! thanks for catching that :)

lib/src/rpc/dial.dart
```diff
@@ -44,6 +46,9 @@ class DialOptions {
   /// If enabled, other auth options have no affect. Eg. [authEntity], [credentials],
   /// [externalAuthAddress], [externalAuthToEntity]
   String? accessToken;
+
+  /// Whether the connection was made usuing MDNS
```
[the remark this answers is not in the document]
clintpurser: great catches, thank you!!

lib/src/rpc/dial.dart
clintpurser: i agree this should be private, I initially didn't do that because goUtils doesn't have it private. But you're right it ought to be. +1 on adding the flag for enabling it, our other implementations also have that, I will add. thank you!

lib/src/rpc/dial.dart
clintpurser: sounds good, had it warning for parody with goUtils, I have no issue with having be debug or info. moving to debug. Thanks!

clintpurser:
@njooma can you take another look at this? 🙏

njooma (approved):

PR #141, opened 2023-10-27
Add new data table

njooma:
Since `MovementSensor` is no longer a sensor type, we can't just reuse the `SensorWidget` like we used to. So I created a new data table to reuse across sensors and movement sensors and anything else that provides a map.

flyby: change the `{Key? key}` to `{super.key}`


benjirewis:
This seems reasonable, but I just don't have the context to approve it 😆 . Gonna remove myself as a reviewer.

clintpurser (approved):

PR #142, opened 2023-10-30
RSDK-5561 - add power sensor screen

clintpurser:
Adds power sensor screen. Also flyby readability change to clean up importing the screens.


clintpurser:
IMG_2896

njooma (approved):
nice

PR #153, opened 2023-11-30
APP-3318 - Fix sticky motor slider

clintpurser:
On Plus/Max sized iphones, the call to set state after awaiting setPower would happen sometimes after the call to stop had gone through, causing the slider to stick after release, even when the toggle to stop after release was enabled.

Moving the setState to before the awaited method calls fixes this issue.

I considered making the calls to setPower and stop `unawaited()` - but then we'd lose our error handling from the try catch.

We could also use the older convention of not using async/await and instead use the Future.onError() callback, to populate the error object. But I opted for using async and await because they are more idiomatic.

### DIFF
--- a/lib/widgets/resources/motor.dart
+++ b/lib/widgets/resources/motor.dart
@@ -36,20 +36,21 @@ class _ViamMotorWidgetState extends State<ViamMotorWidget> {
 
   Future<void> stop() async {
     try {
-      unawaited(widget.motor.setPower(0));
       setState(() {
         power = 0;
       });
+      await widget.motor.stop();
+
       // Sometimes the motor does not honor the first Stop call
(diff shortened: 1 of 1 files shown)


benjirewis (approved):
I swear I've seen this "`setPower` interacts weirdly with `setState(() { this.power = power });`" in one of @njooma's old reviews. Don't have much context here but seems fine to me.

clintpurser:
> I swear I've seen this "`setPower` interacts weirdly with `setState(() { this.power = power });`" in one of @njooma's old reviews. Don't have much context here but seems fine to me.

You're right @benjirewis it does sound familiar. I hope we're not punting different bugs back and forth by changing the order around.

PR #157, opened 2023-12-04
APP-3213 - add wrappers for org members/invites

clintpurser:
adding some wrappers to the app client for organization settings work.

List Organization Members
Create Organization Invites
Delete Organization Invite
Delete Organization Member

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -109,4 +109,34 @@ class AppClient {
         .map((e) => Permission.values.firstWhere((element) => element.value == e))
         .toList();
   }
+
+  Future<ListOrganizationMembersResponse> listOrganizationMembers(Organization org) async {
+    final request = ListOrganizationMembersRequest()..organizationId = org.id;
+    final response = await _client.listOrganizationMembers(request);
+    return response;
+  }
+
+  Future<CreateOrganizationInviteResponse> createOrganizationInvite(Organization org, String email) async {
+    final request = CreateOrganizationInviteRequest()
+      ..organizationId = org.id
+      ..email = email;
+    final response = await _client.createOrganizationInvite(request);
+    return response;
+  }
+
+  Future<DeleteOrganizationInviteResponse> deleteOrganizationInvite(Organization org, String email) async {
+    final request = DeleteOrganizationInviteRequest()
+      ..organizationId = org.id
+      ..email = email;
+    final response = await _client.deleteOrganizationInvite(request);
+    return response;
+  }
+
+  Future<DeleteOrganizationMemberResponse> deleteOrganizationMember(Organization org, String userId) async {
+    final request = DeleteOrganizationMemberRequest()
+      ..organizationId = org.id
+      ..userId = userId;
+    final response = await _client.deleteOrganizationMember(request);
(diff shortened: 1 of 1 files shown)


lib/src/app/app.dart
```diff
@@ -109,4 +109,34 @@ class AppClient {
         .map((e) => Permission.values.firstWhere((element) => element.value == e))
         .toList();
   }
+
+  Future<ListOrganizationMembersResponse> listOrganizationMembers(Organization org) async {
```
clintpurser, line 113: i dislike having the proto definition be the return type here, but otherwise i'd be creating a new wrapper which is essentially the same thing. or alternatively two methods: `List<OrganizationMembers> listOrganizationMembers` & `List<OrganizationInvites> listOrganizationInvites` but that doesn't match the proto. I didn't like either of those ideas more than simply just returning the proto type. Open to changing the approach here though.

lib/src/app/app.dart
```diff
@@ -109,4 +109,34 @@ class AppClient {
...
+
+  Future<OrganizationInvite> createOrganizationInvite(Organization org, String email) async {
+    final request = CreateOrganizationInviteRequest()
+      ..organizationId = org.id
+      ..email = email;
+    final response = await _client.createOrganizationInvite(request);
+    return response.invite;
+  }
+
+  Future<DeleteOrganizationInviteResponse> deleteOrganizationInvite(Organization org, String email) async {
```
clintpurser, line 127: instead of `DeleteOrganizationInviteResponse` I could return `void` - but that makes testing weird. Open to suggestions.

Same thing for `deleteOrganizationMember`'s return type below.

lib/src/app/app.dart
njooma: Ya I think this is fine. I think reading protos is actually not that big of a deal (creating protos is more annoying, but we abstract that away from the user)

lib/src/app/app.dart
njooma: The fact that it's empty makes me think we should actually make this `void`, since having a return type could be confusing

njooma: For testing, instead of the condition being equality on the return type, can we test that the grpc function was called only once?

clintpurser: yeah, thats a good idea, thank you! ✅

lib/src/app/app.dart
clintpurser: sweet. thanks!

njooma (approved):
Nice

lib/src/app/permissions.dart
```diff
@@ -37,3 +51,45 @@ enum Permission {
     return name.replaceAllMapped(exp, (m) => '_${m.group(0)}').toLowerCase();
   }
 }
+
+class ViamAuthorization {
+  String authorizationType;
+
+  /// the authorization itself, either [AuthorizationId.organizationOwner] or [AuthorizationId.organizationOperator]
```
njooma, line 58: [nit] can we make all these comments sentence case?

clintpurser: Good call out, thanks! ✅

PR #160, opened 2023-12-13
improve webrtc connection times & misc improvements

clintpurser:
Quite a bit of misc things going on in this PR - I will try and call them out.

Upgrading flutter_webrtc to latest because: https://github.com/flutter-webrtc/flutter-webrtc/issues/1428

upgrading bonsoir to latest because: https://github.com/Skyost/Bonsoir/issues/57 has been resolved

shorten mDNS scan timeout (this was the main thing lengthening connection times, whoops!).

add try catch for camera stream exception i kept hitting.

add comments for .env setup

### DIFF
--- a/example/viam_example_app/ios/Runner/Info.plist
+++ b/example/viam_example_app/ios/Runner/Info.plist
@@ -25,11 +25,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(FLUTTER_BUILD_NAME)</string>
+	<string>1.0.0</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(FLUTTER_BUILD_NUMBER)</string>
+	<string>1.0.0</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>UILaunchStoryboardName</key>
--- a/example/viam_example_app/lib/login_screen.dart
+++ b/example/viam_example_app/lib/login_screen.dart
@@ -36,8 +36,8 @@ class _LoginScreenState extends State<LoginScreen> {
   /// This method will get called when the widget initializes its state.
   /// It exists outside the overridden [initState] function since it's async.
   Future<void> _initState() async {
-    // Load the `.env` file TODO replace `filename` with your .env filename
-    await dotenv.load(fileName: 'YOUR_ENV_FILE.env');
+    // Load the `.env` file
+    await dotenv.load();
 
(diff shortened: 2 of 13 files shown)


example/viam_example_app/ios/Runner/Info.plist
```diff
@@ -25,11 +25,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(FLUTTER_BUILD_NAME)</string>
+	<string>1.0.0</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(FLUTTER_BUILD_NUMBER)</string>
+	<string>1.0.0</string>
```
clintpurser, line 32: With Flutter 3.16 they changed how they label these strings in the apple projects code - we ought to regenerate our project to address this - but for now this is a fine bandaid.

if people copied this code to start there own apps it could eventually cause problems like this.

example/viam_example_app/lib/login_screen.dart
```diff
@@ -36,8 +36,8 @@ class _LoginScreenState extends State<LoginScreen> {
   /// This method will get called when the widget initializes its state.
   /// It exists outside the overridden [initState] function since it's async.
   Future<void> _initState() async {
-    // Load the `.env` file
-    await dotenv.load();
+    // Load the `.env` file TODO replace `filename` with your .env filename
+    await dotenv.load(fileName: 'YOUR_ENV_FILE.env');
```
njooma, line 40: I think requiring a name for the file is a bit more complicated than just using `.env`. Especially because then they also have to update the `pubspec.yaml` to include their `.env` file (instead of using the default)

example/viam_robot_example_app/lib/main.dart
```diff
@@ -8,7 +8,8 @@ import 'package:viam_sdk/widgets.dart';
 import 'screens/screens.dart';
 
 void main() async {
-  await dotenv.load();
+  // Load the `.env` file TODO replace `filename` with your .env filename
+  await dotenv.load(fileName: 'YOUR_ENV_FILE.env');
```
njooma, line 12: Same comment as the other example `.env.`

njooma (commented):
Thanks! Some comments we can discuss before final review

example/viam_example_app/lib/login_screen.dart
clintpurser: hmmm i think i am doing this differently than you intended then. I didn't realize you could leave it as the default. You're totally right.

njooma (approved):

PR #162, opened 2023-12-29
Refactor robot example

clintpurser:
I was looking into our example code and noticed some quick improvements.

The largest change is pulling out some packages we used in the robot example - this simplifies the code and we're only using vanilla flutter widgets now so it should be more clear to the average reader what is unique to our package and what is not.

Was looking into a mDNS dial failure i was getting and realized we were missing an `await` in the mDNS flow.

Also got our Podfile all pointing to the same iOS minimum version & included mDNS entitlements in the robot example info.plist.

tested locally and both examples still compile and run for me.

### DIFF
--- a/example/viam_robot_example_app/lib/main.dart
+++ b/example/viam_robot_example_app/lib/main.dart
@@ -1,7 +1,5 @@
-import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_dotenv/flutter_dotenv.dart';
-import 'package:flutter_platform_widgets/flutter_platform_widgets.dart';
 import 'package:viam_sdk/viam_sdk.dart';
 import 'package:viam_sdk/widgets.dart';
 
@@ -10,6 +8,7 @@ import 'screens/screens.dart';
 void main() async {
   // Load the `.env` file
   await dotenv.load();
+
   runApp(const MyApp());
 }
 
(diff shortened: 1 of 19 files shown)


example/viam_robot_example_app/lib/main.dart
```diff
@@ -91,42 +54,29 @@ class _MyHomePageState extends State<MyHomePage> {
...
-      robotFut = RobotClient.atAddress(
-        dotenv.env['ROBOT_LOCATION'] ?? '', // or whatever default value you want
-        RobotClientOptions.withApiKey(
-          dotenv.env['API_KEY_ID'] ?? '',
-          dotenv.env['API_KEY'] ?? '',
-        ),
-      );
-    } else {
-      throw Exception('None of the required variables are defined in .env. Please see README.md for more information.');
-    }
```
njooma, line 111: I think maybe we should keep this in, as it allows a user to interact using either the API Key or the Location Secret. Unless we want to deprecate the use of location secrets entirely?

example/viam_robot_example_app/lib/main.dart
```diff
@@ -223,7 +172,7 @@ class _MyHomePageState extends State<MyHomePage> {
                           Text(_robot.resourceNames.where((element) => element.type == resourceTypeComponent).join('\n')),
                         ])
                       : _loading
-                          ? PlatformCircularProgressIndicator()
+                          ? const CircularProgressIndicator()
```
njooma, line 175: Perhaps use `CircularProgressIndicator.adaptive()` to keep it more native (since you're removing all the platform native stuff)

njooma (approved):
Minor comments, but LGTM overall

example/viam_robot_example_app/lib/main.dart
clintpurser: From my understanding all the examples of using Location Secrets have been removed from the App side, I think it makes sense to remove from SDKs too.

example/viam_robot_example_app/lib/main.dart
clintpurser: Ok i'll add that. Just trying to simplify the examples as much as possible. `.adaptive()` on any standard flutter widgets doesn't feel to intrusive.

PR #165, opened 2024-01-03
Add image data api

clintpurser:
After some conversations with @mcvella I realized it was really difficult for SDK consumers to get images from cameras on Viam and use it easily in a Flutter UI. We had already presented this code in our examples, but this PR proposes pulling it into the SDK as a convenience function so that SDK consumers may easily get image data in a format that then easily allows them to use it in a Flutter UI.

I think I added the field in the correct layer, as a function on `Camera` - this did cause some extra overrides, in the tests for example, to need to be added. I don't think adding it only in the `CameraClient` is the right spot, correct me if i'm wrong though because I second guessed myself.

Also, the tests don't feel great. because on paper i don't know what happens under the hood to get the new expected value, but at least we'll know if something changes down the road.

This is also a useful change for the mobile app because Eliot just requested adding live or manual, 1 second,..., 30 second video refresh frequency (like we have in the web app), so this change makes requests like that much easier. Alternative is duplicating the code in that was in the stream.dart file in the mobile app, but I preferred making a function to call in both places instead.

open to feedback here because I feel this is a weird change, could use some guidance on how we want to handle convenience wrappers in our SDKs, i know we do it in other SDKs (python?) but couldn't remember or quickly find any examples in our flutter SDK already.

### DIFF
--- a/lib/src/media/image.dart
+++ b/lib/src/media/image.dart
@@ -89,7 +89,7 @@ class MimeType {
   }
 }
 
-/// A custom image type that contains the [MimeTYpe], raw image data, and lazily loads and caches an [img.Image].
+/// A custom image type that contains the [MimeType], raw image data, and lazily loads and caches an [img.Image].
 class ViamImage {
   /// The mimetype of the image
   final MimeType mimeType;
--- a/test/unit_test/components/camera_test.dart
+++ b/test/unit_test/components/camera_test.dart
@@ -1,3 +1,5 @@
+import 'dart:typed_data';
+
 import 'package:flutter_test/flutter_test.dart';
 import 'package:grpc/grpc.dart';
 import 'package:viam_sdk/src/components/camera/service.dart';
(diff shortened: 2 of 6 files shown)


lib/src/media/image.dart
```diff
@@ -89,7 +89,7 @@ class MimeType {
   }
 }
 
-/// A custom image type that contains the [MimeTYpe], raw image data, and lazily loads and caches an [img.Image].
+/// A custom image type that contains the [MimeType], raw image data, and lazily loads and caches an [img.Image].
```
clintpurser, line 92: flyby: fixed this typo

test/unit_test/components/camera_test.dart
```diff
@@ -8,6 +10,49 @@ import 'package:viam_sdk/viam_sdk.dart';
...
+  98, 114, 209, 10, 22, 36, 52, 225, 37, 241, 23, 24, 25, 26, 38, 39, 40, 41, 42, 53, 54, 55,
+  56, 57, 58, 67, 68, 69, 70, 71, 72, 73, 74, 83, 84, 85, 86, 87, 88, 89, 90, 99, 100, 101, 102,
+  103, 104, 105, 106, 115, 116, 117, 118, 119, 120, 121, 122, 130, 131, 132, 133, 134, 135, 136,
+  137, 138, 146, 147, 148, 149, 150, 151, 152, 153, 154, 162, 163, 164, 165, 166, 167, 168, 169,
+  170, 178, 179, 180, 181, 182, 183, 184, 185, 186, 194, 195, 196, 197, 198, 199, 200, 201, 202,
+  210, 211, 212, 213, 214, 215, 216, 217, 218, 226, 227, 228, 229, 230, 231, 232, 233, 234, 242,
+  243, 244, 245, 246, 247, 248, 249, 250, 255, 218, 0, 12, 3, 1, 0, 2, 17, 3, 17, 0, 63, 0, 254,
+  178, 40, 0, 255, 217 // prevent dartfmt
+];
+
```
clintpurser, line 55: i almost pulled this into its own file (please give me a reason to haha) - I tried to make this as small as possible, but getting a valid JPEG in a List<int> format was difficult.

writing this comment I'm realizing this needs a comment explaining why I have this now here in this format. I will add that....

TLDR, testing with the [0,0,0] started breaking the converters inside the imageData() method and throwing exceptions from the Image package. so in order to test the new helper method, we have to use bytes for a valid image, i thought the smallest would be a valid jpeg which i ran a script to convert into bytes and then saved here.

njooma:
My thoughts are that this shouldn't be in the dart part of the SDK, but rather in the flutter/widget part. We should make a custom widget that can display a `ViamImage`, and also a helper function contained within an extension on `ViamImage` to get a `ui.Image`.

I agree that it's super gross converting things from `bytes` to `img.Image` to `ui.Image` to `UInt8List`. So let's take it all away -- give them a default widget they can use, and also give them functions in case they want to do their own thing.

clintpurser:
@njooma  - I agree we shouldn't add any more flutter dependencies in the dart part of the SDK - I didn't realize i had pulled in one, but I was able to remove it. 

I think providing the imageData wrapper is still a useful way to provide the data to SDK consumers, because some will want more finite control over their state and a pre wrapped widget doesn't always handle things how you want, but I also agree I see the value of providing the widgets that handle everything for the user.

One example of needing more finite control that immediately comes to mind is how could we link a button for manually refreshing the image - like we have in the web app. I am not thinking of an easy way to do this only with widgets and not using state and managing the data outside of the build method.

I've added some wrapped widgets, which use the new method to get their data, but still left the new function in.

https://github.com/viamrobotics/viam-flutter-sdk/assets/22106496/724f1006-e7ad-4385-9cf3-d836d15f9c96

stuqdog:
Hiya @clintpurser! Just fyi I'm removing myself as a reviewer on this so it will no longer appear in my review pile. Feel free to re-add me if this leaves draft stage at some point :)

PR #176, opened 2024-01-29
RSDK-5953 - add newMachine wrapper

clintpurser:
wrapping the newRobot call with `newMachine`.

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -164,4 +164,13 @@ class AppClient {
       ..userId = userId;
     await _client.deleteOrganizationMember(request);
   }
+
+  /// Create a new smart machine with the included [name] in the passed in [locationId]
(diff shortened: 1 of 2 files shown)


lib/src/app/app.dart
```diff
@@ -164,4 +164,13 @@ class AppClient {
       ..userId = userId;
     await _client.deleteOrganizationMember(request);
   }
+
+  /// Create a new smart machine with the included [name] in the passed in [locationId]
+  Future<String> newMachine(String name, String locationId) async {
```
clintpurser, line 169: debated calling this `newRobot` to be more transparent with the underlying proto - but it seems like that would be changing soonish and would become a breaking change. Trying to mitigate the amount of those in the future. But I can also see the argument with staying consistent and doing a hard cutover in the future.

PR #179, opened 2024-01-31
APP-3811 - Provisioning wrappers

clintpurser:
This is waiting on https://github.com/viamrobotics/api/pull/438

Add client wrappers and tests for the new provisioning API.

In its current state, this PR also adds generated code for provisioning that I manually generated from the above linked PR/branch. This won't actually get merged, but it is included here so I could write the wrappers and tests - and other work in the mobile app while I wait for that work to get finished.


clintpurser:
@njooma i marked this as draft because we'll want to wait for the API change to come through first. But could you take a look at the client and the tests in the meantime? Thanks!

lib/src/app/provisioning.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// Get the status of the SmartMachine
+  Future<GetSmartMachineStatusResponse> getSmartMachineStatus() async {
+    final request = GetSmartMachineStatusRequest();
+    final response = await _client.getSmartMachineStatus(request);
+    return response;
+  }
+
+  /// Set the network credentials of the Smart machine, so it can connect to the internet.
+  ///
+  /// [type] is either 'wired' or 'wifi'
```
njooma, line 22: Maybe make this an `enum` to make it strongly typed?

njooma (commented):
Great start. Do you not have to also add this to the list of library exports though?

lib/src/app/provisioning.dart
clintpurser: good point, i thought about that but skipped for now. Will circle back and add in. Thanks!

clintpurser:
> Great start. Do you not have to also add this to the list of library exports though?

ah yup, totally forgot that. Great catch, thanks!

lib/src/app/provisioning.dart
```diff
@@ -0,0 +1,52 @@
...
+    final request = GetSmartMachineStatusRequest();
+    final response = await _client.getSmartMachineStatus(request);
+    return response;
+  }
+
+  /// Set the network credentials of the Smart machine, so it can connect to the internet.
+  ///
+  /// [ssid] is the networks ssid
+  /// [psk] is the networks passkey
+  Future<void> setNetworkCredentials({NetworkType type = NetworkType.wifi, required String ssid, required String psk}) async {
```
njooma, line 26: `ssid` and `psk` shouldn't be required for `wired` connections. Not a locker, but please add a TODO

lib/src/app/provisioning.dart
```diff
@@ -0,0 +1,52 @@
...
+    final request = SetNetworkCredentialsRequest(type: type.name, ssid: ssid, psk: psk);
+    await _client.setNetworkCredentials(request);
+  }
+
+  /// Set the Viam credentials of the Smart machine, so it can connect to the Viam.
```
njooma, line 31: > connect to the Viam

😂

njooma (approved):

PR #185, opened 2024-03-11
RSDK-6116: Mobile app should auto refresh board's analog readings

Oabraham1:
[Ticket](https://viam.atlassian.net/browse/RSDK-6116)

In this PR, I modified the board card to refresh board readings every second just like we do for the sensor readings.

### DIFF
--- a/lib/widgets/resources/board.dart
+++ b/lib/widgets/resources/board.dart
@@ -13,10 +13,12 @@ import '../button.dart';
 class ViamBoardWidget extends StatefulWidget {
   /// The [Board]
   final Board board;
+  final Duration refreshInterval;
 
   const ViamBoardWidget({
     super.key,
     required this.board,
+    this.refreshInterval = const Duration(seconds: 1),
   });
 
   @override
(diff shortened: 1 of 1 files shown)


njooma (changes requested):
The timer needs to be invalidated on destruction of this widget

stuqdog (approved):
looks reasonable to me

lib/widgets/resources/board.dart
```diff
@@ -34,6 +36,9 @@ class _ViamBoardWidgetState extends State<ViamBoardWidget> {
   final _pwmFreqFormKey = GlobalKey<FormState>();
   int pwmFrequency = 0;
 
+  Duration refreshInterval = const Duration(seconds: 1);
```
njooma, line 39: What do you think about making this a property of the widget? That way a consumer could specify the refresh interval themself?

Oabraham1: DO you think it would be valuable to default it to 1 second though?

lib/widgets/resources/board.dart
```diff
@@ -11,10 +13,12 @@ import '../button.dart';
 class ViamBoardWidget extends StatefulWidget {
   /// The [Board]
   final Board board;
+  final Duration refreshInterval;
```
njooma, line 16: Add a docstring comment describing this property please

njooma (changes requested):

lib/widgets/resources/board.dart
Oabraham1: Added

njooma (approved):

PR #189, opened 2024-04-03
RSDK-7045 Implement StreamTicks 

oliviamiller:
Implements the StreamTicks board API. also tested with code samples client script.

### DIFF
--- a/lib/src/components/board/board.dart
+++ b/lib/src/components/board/board.dart
@@ -1,9 +1,10 @@
 import 'dart:collection';
 
 import 'package:fixnum/fixnum.dart';
+import 'package:grpc/grpc.dart';
 
 import '../../gen/common/v1/common.pb.dart' as common;
-import '../../gen/component/board/v1/board.pbenum.dart';
+import '../../gen/component/board/v1/board.pbgrpc.dart';
 import '../../resource/base.dart';
 import '../../robot/client.dart';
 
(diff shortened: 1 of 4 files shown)


lib/src/components/board/board.dart
```diff
@@ -58,6 +70,12 @@ abstract class Board extends Resource {
   /// Return the current value of the interrupt which is based on the type of Interrupt.
   Future<int> digitalInterruptValue(String digitalInterruptName, {Map<String, dynamic>? extra});
 
+  // Stream digital interrupts ticks.
+  Future<ResponseStream<StreamTicksResponse>> streamTicks(List<String> interrupts, {Map<String, dynamic>? extra});
```
njooma, line 74: Can we make the return type a `Stream` of `Tick`s please. Take a look at how we do it for [`tailLogs`](https://github.com/viamrobotics/viam-flutter-sdk/blob/337e9d2767eec68c022bca94a6f3c852b94b822b/lib/src/app/app.dart#L95) to get an idea of how to convert a `ResponseStream` to a native `Stream`

lib/src/components/board/service.dart
```diff
@@ -112,8 +114,23 @@ class BoardService extends BoardServiceBase {
...
-    // TODO: implement streamTicks
-    throw UnimplementedError();
+  Stream<StreamTicksResponse> streamTicks(ServiceCall call, StreamTicksRequest request) async* {
+    final board = _fromManager(request.name);
+
+    final ticks = Queue<Tick>();
+    await board.addCallbacks(request.pinNames, ticks);
+
+    try {
+      while (true) {
```
njooma, line 124: Does this ever end? I don't see a break ever... Not sure if it's supposed to end ever but just wondering

njooma (changes requested):

lib/src/components/board/service.dart
oliviamiller: No its supposed to continue forever until the client cancels it.

lib/src/components/board/client.dart
```diff
@@ -120,6 +123,22 @@ class BoardClient extends Board implements ResourceRPCClient {
...
+  Future<void> addCallbacks(List<String> interrupts, Queue<Tick> tickQueue, {Map<String, dynamic>? extra}) async {
+    // addCallbacks not implemented on client side since it is a helper for StreamTicks.
+    throw UnimplementedError();
+  }
+
+  @override
+  Stream<Tick> streamTicks(List<String> interrupts, {Map<String, dynamic>? extra}) {
+    final response = client.streamTicks(StreamTicksRequest()
+      ..name = name
+      ..pinNames.addAll(interrupts));
```
stuqdog, line 136: looks like we're not passing along `extra` here, is that intentional?

stuqdog (commented):
Generally looks good to me! modulo question about `extra`

lib/src/components/board/client.dart
oliviamiller: not intentional thanks for catching! added

njooma (approved):

stuqdog (approved):

PR #193, opened 2024-04-15
RSDK-7309: Update Status for Board in Flutter SDK

martha-johnston:


martha-johnston:
I'm a bit confused with how the widget stuff works, could anyone more familiar with the mobile setup offer suggestions? I want the analogs and digitals to still update regularly if they exist, but the status func is no longer available to poll for that info, it should operate the way other components do, maybe similar to sensor? not super sure

njooma (approved):

PR #201, opened 2024-05-07
RSDK-7217 - Data wrappers

stuqdog:


test/unit_test/app/data_client_test.dart
```diff
@@ -8,12 +8,78 @@ import 'package:mockito/mockito.dart';
...
 import '../mocks/mock_response_future.dart';
 import '../mocks/service_clients_mocks.mocks.dart';
 
+class FakeDataServiceClient extends MockDataServiceClient {
+  bool tagsAddedByIds = false;
+  bool tagsAddedByFilter = false;
+  bool removeBboxIdCalled = false;
+  bool configureDatabaseUserCalled = false;
+  bool addBinaryDataToDatasetByIdsCalled = false;
+  bool removeBinaryDataFromDatasetByIdsCalled = false;
```
njooma, line 26: You don't need these, you should be able to use `mockito` to confirm that these functions were called: https://pub.dev/packages/mockito#verifying-exact-number-of-invocations--at-least-x--never

njooma: Which i think means that this entire fake client can be removed

njooma (changes requested):

test/unit_test/app/data_client_test.dart
stuqdog: Oh fantastic, thanks!

PR #203, opened 2024-05-09
RSDK-7216: add app wrappers

purplenicole730:


lib/src/app/app.dart
```diff
@@ -16,25 +17,151 @@ class AppClient {
 
   AppClient(this._client);
 
+  /// Get the id of the user with the email provided
+  Future<String> getUserIDByEmail(String email) async {
```
njooma, line 21: [nit] `ID` should be `Id` (see [Dart Style Guide](https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words))

lib/src/app/app.dart
```diff
@@ -92,15 +271,178 @@ class AppClient {
...
+  }
+
+  /// Delete a [RobotPart]
+  Future<void> deleteRobotPart(String partId) async {
+    final request = DeleteRobotPartRequest()..partId = partId;
+    await _client.deleteRobotPart(request);
+  }
+
+  /// Gets the [APIKey]'s for a [Robot]
+  Future<List<APIKeyWithAuthorizations>> getRobotAPIKeys(String robotId) async {
```
njooma, line 317: [nit] should be `getRobotApiKeys` (capital-case API)

njooma (changes requested):
A couple of nits. TBH, didn't read everything it's too much 😭

lib/src/app/app.dart
```diff
@@ -16,25 +17,151 @@ class AppClient {
...
   /// List all the [Organization] the currently authenticated user has access to
   Future<List<Organization>> listOrganizations() async {
     final listOrganizationsRequest = ListOrganizationsRequest();
     final ListOrganizationsResponse response = await _client.listOrganizations(listOrganizationsRequest);
     return response.organizations;
   }
 
+  /// Get all [OrganizationIdentity]s that have access to a [Location].
+  Future<List<OrganizationIdentity>> getOrganizationsWithAccessToLocation(String locationId) async {
+    final request = GetOrganizationsWithAccessToLocationRequest();
```
stuqdog, line 43: We need to set the `locationId` here.

stuqdog: As an aside, it's unfortunate that we don't have a linter to catch unused method params. I wonder if there's an obvious choice for such a thing, or a reason we don't have one set up currently @njooma @clintpurser?

lib/src/app/app.dart
```diff
@@ -16,25 +17,151 @@ class AppClient {
...
+
+  /// Update an [Organization]
+  Future<Organization> updateOrganization(String organizationId,
+      {String? name, String? publicNamespace, String? region, String? cid}) async {
+    final request = UpdateOrganizationRequest()
+      ..organizationId = organizationId
+      ..name = name ?? ''
+      ..publicNamespace = publicNamespace ?? ''
+      ..region = region ?? ''
+      ..cid = cid ?? '';
```
stuqdog, line 77: I _think_ instead we should be doing, e.g., 
```dart
final request = UpdateOrganizationRequest();
if (name != null) request.name = name;
```

My fear is that by setting to an empty string, on the app side these values will get overridden by an empty string (as opposed to left untouched).

lib/src/app/app.dart
```diff
@@ -16,25 +17,151 @@ class AppClient {
...
+
+  /// Delete an [Organization]
+  Future<void> deleteOrganization(String organizationId) async {
+    final request = DeleteOrganizationRequest()..organizationId = organizationId;
+    await _client.deleteOrganization(request);
+  }
+
+  /// List the members and pending invites for an [Organization].
+  Future<ListOrganizationMembersResponse> listOrganizationMembers(Organization org) async {
+    final request = ListOrganizationMembersRequest()..organizationId = org.id;
```
stuqdog, line 90: We only care about the `org` argument here for its id. Passing a string is a lot less work than creating and passing a whole `Organization` object, can the param just be a string instead?

lib/src/app/app.dart
```diff
@@ -16,25 +17,151 @@ class AppClient {
...
+    final ListOrganizationMembersResponse response = await _client.listOrganizationMembers(request);
+    return response;
+  }
+
+  /// Send an invitation to to join an [Organization] to the specified email. Grant the level of permission defined in the [ViamAuthorization] object attached.
+  Future<OrganizationInvite> createOrganizationInvite(Organization org, String email, List<ViamAuthorization> authorizations,
```
stuqdog, line 96: See above re: `org` param versus `orgId` param.

edit: it looks like this happens in a number of cases below as well. Can we replace all insances of an `org` or `location` parameter where we just grab an orgId or locationId with an `orgId` or `locationId` parameter?

lib/src/app/app.dart
```diff
@@ -44,11 +171,66 @@ class AppClient {
...
-    final listRobotsRequest = ListRobotsRequest()..locationId = location.id;
-    final ListRobotsResponse response = await _client.listRobots(listRobotsRequest);
-    return response.robots;
+  /// Update a [Location]
+  Future<Location> updateLocation(String locationId, {String? name, String? parentLocationId, String? region}) async {
+    final request = UpdateLocationRequest()
+      ..locationId = locationId
+      ..name = name ?? ''
+      ..parentLocationId = parentLocationId ?? ''
+      ..region = region ?? '';
```
stuqdog, line 180: I think, similar to `updateOrganization`, we want to not set these values if they're not provided.

lib/src/app/app.dart
```diff
@@ -92,15 +271,178 @@ class AppClient {
   }
 
   /// Get a stream of [LogEntry] for a specific [RobotPart]. Logs are sorted by descending time (newest first)
-  Stream<List<LogEntry>> tailLogs(RobotPart part, {bool errorsOnly = false}) {
+  Stream<List<LogEntry>> tailLogs(RobotPart part, {bool errorsOnly = false, String? filter}) {
```
stuqdog, line 274: `part` should probably just be `String partId`

lib/src/app/app.dart
```diff
@@ -92,15 +271,178 @@ class AppClient {
...
 
+  /// Get a specific [RobotPart] history by ID
+  Future<List<RobotPartHistoryEntry>> getRobotPartHistory(String id) async {
+    final request = GetRobotPartHistoryRequest()..id = id;
+    final GetRobotPartHistoryResponse response = await _client.getRobotPartHistory(request);
+    return response.history;
+  }
+
+  /// Update a specific [RobotPart] by ID
+  Future<RobotPart> updateRobotPart(String partId, String name, Struct robotConfig) async {
```
stuqdog, line 292: It looks like the idiom elsewhere in the SDK is to ask people to pass `struct`-like objects as a `Map<String, dynamic>` and then convert to a struct under the hood. Can we change to doing so here (and elsewhere where we ask for a `Struct` param)?

test/unit_test/app/app_client_test.dart
```diff
@@ -167,24 +432,326 @@ void main() {
...
+    test('getRegistryItem', () async {
+      final expected = RegistryItem()
+        ..organizationId = 'organizationId'
+        ..name = 'name'
+        ..type = PackageType.PACKAGE_TYPE_UNSPECIFIED;
+      when(serviceClient.getRegistryItem(any)).thenAnswer((_) => MockResponseFuture.value(GetRegistryItemResponse()..item = expected));
+      final response = await appClient.getRegistryItem('itemId');
+      expect(response, equals(expected));
+    });
+    test('createRegistryItem', () async {
```
stuqdog, line 622: (minor) line break between tests

stuqdog (changes requested):
Nice, generally looks good. Just some suggestions around method params (using, setting, making more ergonomic).

lib/src/app/app.dart
njooma: I definitely started this back in the day. I agree that passing IDs is the easiest way to do it and should probably be switched

lib/src/app/app.dart
njooma: There is no linter option for this: https://github.com/dart-lang/linter/issues/2204

lib/src/app/app.dart
purplenicole730: That was an original function. I didn't change this.

lib/src/app/app.dart
purplenicole730: Yep, will do. These were the original functions that were already in the client.

lib/src/app/app.dart
purplenicole730: Similar to my previous response lol

lib/src/app/app.dart
stuqdog: 😭 so it goes!

lib/src/app/app.dart
stuqdog: Ahh okay, that makes sense! I didn't realize 😅 I think if we can make the change easily that's a great flyby :)

stuqdog (approved):
Nice, lgtm!

purplenicole730:
> Nice, lgtm!

Thanks so much for looking through this entire PR! I really appreciate it since my eyes/brain were fried by then time I finished 😭

njooma (approved):

PR #210, opened 2024-05-13
RSDK-4264 - Move to api gostream dependency

stuqdog:


njooma (approved):
I assume streaming video was tested with the new protos

stuqdog:
> I assume streaming video was tested with the new protos

Ahh good question! I did not test that, though I did `vimdiff` the old and new protos and confirmed that there were no differences. I tried testing streaming video but am having some issues getting the rc-app to run locally right now... will continue to dig and confirm testing tomorrow.

purplenicole730 (approved):

PR #215, opened 2024-05-16
return new analog reader response

martha-johnston:


stuqdog (changes requested):
The code generally looks good! a couple small changes requested but that's it.

However, we should probably be forking this off of the `workflows/update-protos` branch rather than `main`, so that all the proto updating happens in a consistent reliable place.

martha-johnston:
> However, we should probably be forking this off of the `workflows/update-protos` branch rather than `main`, so that all the proto updating happens in a consistent reliable place.

sure I'll switch my fork for future PRs

stuqdog:
> > However, we should probably be forking this off of the `workflows/update-protos` branch rather than `main`, so that all the proto updating happens in a consistent reliable place.
> 
> sure I'll switch my fork for future PRs. is there anything I need to do in this one to make sure it's up to date?

We should update so that we're merging into `workflows/update-protos` rather than `main`, then merge this into `update-protos` and merge that into `main`.

stuqdog (approved):
lgtm (modulo the discussed branch fixes).

PR #223, opened 2024-06-05
Fix pwmFrequency and setPowerMode examples

JessamyT:


purplenicole730 (approved):

purplenicole730 (approved):

lib/src/components/board/board.dart
```diff
@@ -100,7 +101,10 @@ abstract class Board extends Resource {
   ///
   /// ```
   /// // Set the power mode of the board to offline deep for 60 seconds
-  /// await myBoard.setPowerMode(POWER_MODE_OFFLINE_DEEP, 60, 0);
+  /// // Requires adding 'package:viam_sdk/protos/component/board.dart'
+  /// // to your imports
```
njooma, line 105: Probably easier to just say `Requires importing `package:...` and put it all one one line

njooma (approved):
Small nit but otherwise good

lib/src/components/board/board.dart
JessamyT: Ahh good call ty
```suggestion
  /// // Requires importing 'package:viam_sdk/protos/component/board.dart'
```

PR #224, opened 2024-06-05
`viam_example_app` has syntax's errors

g0rdan:
I'm unsure if there is a code of conduct or a PR template for PRs but I'll post it as is.

I wanted to look at the SDK and I found the example app, most likely not up-to-date with the latest changes in the SDK.
So, here is a possible fix for it.


njooma (approved):
Thanks for updating the example app!

PR #231, opened 2024-06-13
DOCS-2400: Add snippets for resource base methods

JessamyT:


lib/src/resource/base.dart
```diff
@@ -26,12 +26,22 @@ class Subtype {
   const Subtype(this.namespace, this.resourceType, this.resourceSubtype);
 
   /// Create a new [Subtype] from a [ResourceName]
+  ///
+  /// ```
+  /// // Example, creating a Sensor from a resource named "mySensor":
+  /// var mySensor = Sensor.fromResourceName('mySensor');
+  /// ```
```
JessamyT, line 33:
```suggestion
```
Should these actually not be here since they're not part of the API?

lib/src/resource/base.dart
```diff
@@ -26,12 +26,22 @@ class Subtype {
...
       : namespace = name.namespace,
         resourceType = name.type,
         resourceSubtype = name.subtype;
 
   /// Get a [ResourceName] from this [Subtype] and a provided [name]
+  ///
+  /// ```
+  /// // Example, getting the ResourceName of a sensor named "my_sensor":
+  /// var resourceName = Sensor.getResourceName('my_sensor');
+  /// ```
```
JessamyT, line 44:
```suggestion
```

lib/src/resource/base.dart
stuqdog: There may be considerations I'm not thinking of but it seems harmless and more thorough to include these.

stuqdog (approved):

PR #240, opened 2024-06-28
DOCS-2591: Add motor and servo example snippets

JessamyT:
Also fixes my power sensor mistakes 👀


stuqdog (approved):

JessamyT:
I was awaiting @njooma 's review since Ethan and Naveed are both auto-populated into the reviewers list, but maybe I don't need to?

njooma:
Oops sorry I was waiting to review until all the tests were passing

JessamyT:
Is there something I should do to make the htmltest pass? I was thinking it's okay since it's failing on things that aren't part of the diff but not sure how to fix.

PR #245, opened 2024-07-15
UpdateFragment/ ListFragments - add visibility parameter

piokasar:


lib/src/app/app.dart
```diff
@@ -379,10 +379,11 @@ class AppClient {
   }
 
   /// Get a list of [Fragment]s in an [Organization]
-  Future<List<Fragment>> listFragments(String organizationId, bool showPublic) async {
+  Future<List<Fragment>> listFragments(String organizationId, {bool? showPublic}, {FragmentVisibility? visibility}) async {
```
piokasar, line 382: changed this to be optional given that it's an optional field in api - let me know if that seems wrong

lib/src/app/app.dart
```diff
@@ -471,8 +471,8 @@ class AppClient {
   /// Get a list of [Fragment]s in an [Organization]
   ///
   /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
-  Future<List<Fragment>> listFragments(String organizationId, bool showPublic) async {
-    final request = ListFragmentsRequest()
+  Future<List<Fragment>> listFragments(String organizationId, bool showPublic, {List<FragmentVisibility>? fragmentVisibility}) async {
+    final request = ListFragmentsRequest(fragmentVisibility: fragmentVisibility)
```
njooma, line 475: Can you please switch this to the `..fragmentVisiblity = fragmentVisibility` syntax like the other options?

njooma (changes requested):

lib/src/app/app.dart
piokasar: changed !

piokasar: ok, actually should be good to go now - had to use addAll since it's an iterable

purplenicole730 (approved):

njooma (approved):

PR #251, opened 2024-08-01
RSDK-8304 - send version metadata

stuqdog:
How it works:
 - `update-protos.yml` updates the `apiTag` value in the `getVersionMetadata` function
 - `release.yml` updates the `sdkVersion` value in the `getVersionMetadata` function
 - channels update their options to set `viam_client` equal to the output of `getVersionMetadata`

Note that this is a bit of a bummer because we're storing the SDK version in two places (`pubspec.yml` and `utils.dart`). I spent a good amount of time trying to find some alternative whereby we'd get the version info out of `pubspec.yml` instead. The potential solutions and (seemingly intractable) problems I encountered were:
 - you can't read data out of `pubspec.yml` directly in a synchronous way, which means that reading `pubspec.yml` data would require polluting our API with async, which would be a significant breaking change and open tons of new cans of worms.
 - Though it's not recommended, you _can_ just open a file synchronously (i.e., instead of saying "give me pubspec data" we could say "open the file found at `pubspec.yml` and convert it into a map that I can parse). This seemed promising and tests passed, but the problem I found was the absolute path to `pubspec.yml` can't be reliably assumed, and setting a relative path meant the file couldn't always be found (it was found successfully for a unit test, but not by the `flutter-rc-app` running from a sibling directory). So, this approach was too fragile to work.
 - Technically, you can still force an async function in dart to run to completion and block synchronously, meaning we _could_ get the ask for data from the `pubspec.yml` as per the first option above. But, the function to do so is deprecated, will be removed soon, and its use is strongly discouraged. So that approach is neither future proof nor idiomatic.

Given the above, I opted for the simple and consistent, but sadly duplicative, approach that we wound up with here. 

Screenshot 2024-08-01 at 09 27 42

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -26,6 +26,9 @@ jobs:
 
       - run: make setup
 
+      - name: Update Proto Tag
+        run: sed -i -e "s/apiTag = 'v[0-9]*\.[0-9]*\.[0-9]*'/apiTag = '${{ github.event.client_payload.tag }}'" > lib/src/utils.dart
+
       - name: Generate buf
         run: make buf
         env:
@@ -33,9 +36,6 @@ jobs:
 
       - name: Format
         run: make format
-
-      - name: Update Proto Tag
-        run: sed -i -e "s/apiTag = 'v[0-9]*\.[0-9]*\.[0-9]*'/api_tag = '${{ github.event.client_payload.tag }}'" > lib/src/utils.dart
 
       - name: Add + Commit + Open PR
         uses: peter-evans/create-pull-request@v5
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Format
         run: make format
 
+      - name: Update Proto Tag
+        run: sed -i -e "s/apiTag = 'v[0-9]*\.[0-9]*\.[0-9]*'/api_tag = '${{ github.event.client_payload.tag }}'" > lib/src/utils.dart
+
       - name: Add + Commit + Open PR
         uses: peter-evans/create-pull-request@v5
         with:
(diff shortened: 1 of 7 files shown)


lib/src/utils.dart
```diff
@@ -82,3 +83,9 @@ extension MapStructUtils on Map<String, dynamic> {
     return Value()..structValue = toStruct();
   }
 }
+
+String getVersionMetadata() {
+  const String sdkVersion = 'v0.0.18';
+  const String apiTag = 'v0.1.328';
+  return format('flutter;{};{}', sdkVersion, apiTag);
```
njooma, line 90: Don't need the format library -- you can use format strings:

```dart
return 'flutter;$sdkVersion;$apiTag'
```

pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dependencies:
   collection: ^1.17.1
   async: ^2.11.0
   bonsoir: ^5.1.8
+  format: ^1.5.2
```
njooma, line 23: Won't be needed

.github/workflows/update_protos.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       - name: Format
         run: make format
 
+      - name: Update Proto Tag
+        run: sed -i -e "s/apiTag = 'v[0-9]*\.[0-9]*\.[0-9]*'/api_tag = '${{ github.event.client_payload.tag }}'" > lib/src/utils.dart
+
```
njooma, line 39: Can we update the proto tag earlier, and then update the Makefile so that it reads the `lib/src/utils.dart` file and grabs the `apiTag`, and then generates that specific api tag?

njooma (changes requested):
Can we have the `make buf` command grab the `apiTag` from this file?

njooma (approved):

PR #255, opened 2024-08-21
RSDK-8607 - fix timeout generation bug

stuqdog:
Fixes a bug whereby we were doubling timeout durations by putting the full value in both the `seconds` and `nanos` fields, which incidentally meant that we were getting an int overflow error when trying to pass a timeout of ~2sec or greater.


stuqdog:
@kevin49999 fyi

purplenicole730 (approved):

PR #260, opened 2024-09-16
make RobotClientOptions properties *not* final

clintpurser:
Ran into an issue with this where i wanted to set `enableSessions` to false - it was not possible because it was `final` and the value was set upon initialization.

An alternative approach could be leaving them as final, but providing the default values in the constructors.


njooma (approved):
I don't see why we can't let people change their options outside the constructor

PR #261, opened 2024-09-19
RSDK-8812: Add frame_rate to flutter sdk

jckras:
- added support for `frame_rate` by returning it in getProperties in service.dart
- added a test in camera_test.dart to test for setting vs not setting `frame_rate`


njooma (changes requested):
All the tests actually become redundant and you can undo all your changes since we're simply returning the properties values as is.

njooma (approved):

PR #262, opened 2024-09-23
RSDK-7961: highlight viam sdk section

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-7961)

There is now a new topic/category called Viam SDK that will show up in the navigation bar. It will show above the "Protobuf Definitions" topic.
image

### NOTE ABOUT VIAM SDK LIBRARY "NECESSARY PROTOS"
There were a couple of protos defined in the viam sdk library under the comment: "Necessary Protos". These proto classes are in the "Protobuf Definitions" category, so cannot be part of the new category. These still show up in the regular [viam sdk library link](https://flutter.viam.dev/viam_sdk/viam_sdk-library.html), just without the `Viam SDK` tag. These are **NOT** in the Viam SDK topic.
- CloudConfig
- Credentials
- DoCommandRequest
- DoCommandResponse
- GeoPoint
- GetNetworkListRequest
- GetNetworkListResponse
- GetSmartMachineStatusRequest
- GetSmartMachineStatusResponse
- NetworkInfo
- Orientation
- Pose
- PoseInFrame
- ProvisioningInfo
- ResourceName
- SetNetworkCredentialsRequest
- SetNetworkCredentialsResponse
- SetSmartMachineCredentialsRequest
- SetSmartMachineCredentialsResponse
- Vector3

Example showing that the protos are not in the Viam SDK category, but is in the library: image

### DIFF
--- a/doc/ViamSDK.md
+++ b/doc/ViamSDK.md
@@ -1,3 +1,3 @@
 # Viam SDK Library
 
-Robots at your fingertips.
+Smart machines on smart devices.
--- a/doc/ViamSDK.md
+++ b/doc/ViamSDK.md
@@ -1,0 +1,3 @@
+# Viam SDK Library
+
+Robots at your fingertips.
--- a/dartdoc_options.yaml
+++ b/dartdoc_options.yaml
@@ -1,7 +1,10 @@
 dartdoc:
   categories:
+    "Viam SDK":
+      markdown: doc/ViamSDK.md
+      name: Viam SDK
     "Protobuf Definitions":
       markdown: doc/Protobuf.md
       name: Protobuf Definitions
-  categoryOrder: ["Protobuf Definitions"]
-  nodoc: ['**/di/**', '**/domain/**']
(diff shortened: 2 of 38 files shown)


doc/ViamSDK.md
```diff
@@ -0,0 +1,3 @@
+# Viam SDK Library
+
+Robots at your fingertips.
```
npentrel, line 3: Open to alternatives but we shouldn't add more occurances of the term robot
```suggestion
Work with your machines and hardware.
```

njooma: Not an additional occurrence but a copy from the README. 

But you are correct we shouldn't be using robot anymore. This was a pun since the flutter SDK is about connecting your smart phone to machines. I would like it if we could remain punny but obvi don't want to spend too much time on coming up with a new one

Maybe "Smart machines on smart devices"??

njooma (approved):

PR #266, opened 2024-09-25
RSDK-7962: organize sections in docs

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-7962)

There are three new topics, "Components", "Services", and "Common". Component/service classes were moved out of the "Viam SDK" category.
The common library is auto generated, so the `export_protos.dart` file checks for the common library and adds it to the new "Common" topic instead of "Protobuf Definitions" topic. You can only have one category at a time.

Unfortunately, setting the category for specific classes seems to happen at class definition, so the "Common" topic would have to be set in the auto-generated file. Defining the category while exporting common classes (in `lib/viam_sdk.dart`) did not work. I thought it should have, but I'm running into issues during testing this.
There is an alternative, but less ideal, method of bringing attention to specific `common` classes. `Common.md` links the common classes that are specifically in the Viam SDK library.
image

Note: I did try using the `subCategories` to put everything under the Viam SDK category, but that doesn't seem to change the actual visuals based on what I've read and my own testing.


npentrel (approved):
LGTM if y'all are happy with this, thanks!

lib/src/components/arm/service.dart
```diff
@@ -6,6 +6,7 @@ import '../../resource/manager.dart';
 import '../../utils.dart';
 import 'arm.dart';
 
+/// {@category Components}
```
njooma, line 9: All of the category definitions in all of the `service.dart` files are actually unneeded since we don't export these files, so no documentation exists for these

njooma (approved):

lib/src/components/arm/service.dart
njooma: (No need to do anything, just commenting here for documentation)

PR #274, opened 2024-10-08
[RSDK-8920] Add DiscoverComponents to robot client

hexbabe:
https://viam.atlassian.net/browse/RSDK-8920

- [x] Bump proto version to get new DiscoveryQuery field `extra`
- [x] Add method and new class to robot client
- [x] Test with module that implements Discovery

First PR in Flutter SDK and first time writing code Dart; appreciate tips and best practice comments on top of correctness


lib/src/robot/client.dart
```diff
@@ -5,17 +5,19 @@ import 'package:grpc/grpc_connection_interface.dart';
 import 'package:logger/logger.dart';
 
 import '../gen/common/v1/common.pb.dart';
-import '../gen/robot/v1/robot.pbgrpc.dart';
```
hexbabe, line 8: "Real" non-proto update changes occur in this file

lib/src/robot/client.dart
```diff
@@ -5,17 +5,19 @@ import 'package:grpc/grpc_connection_interface.dart';
...
+import '../gen/google/protobuf/struct.pb.dart';
+import '../gen/robot/v1/robot.pbgrpc.dart' as rpb;
```
hexbabe, line 9: import aliasing robot pb to avoid identifier collision between the new `DiscoveryQuery` in the SDK, and the pb `DiscoveryQuery`

lib/src/robot/client.dart
```diff
@@ -52,6 +54,18 @@ class RobotClientOptions {
...
+class DiscoveryQuery {
+  final String subtype;
+  final String model;
+  final Map<String, dynamic> extra;
+
+  DiscoveryQuery({required this.subtype, required this.model, Map<String, dynamic>? extra}) : extra = extra ?? {};
+
+  Struct get extraStruct => extra.toStruct();
+}
+
```
njooma, line 68: I don't think this class should exist and that people should just use the `DiscoveryQuery` proto directly. This doesn't add any additional functionality or ergonomics to the developer IMO

njooma (changes requested):

lib/src/robot/client.dart
hexbabe: If we don't make this class, how do we import `rpb`'s proto `DiscoveryQuery` in our app? I tried using `import 'package:viam_sdk/viam_sdk.dart';` and it doesn't seem to import proto logic.

hexbabe:
```
lib/main.dart:53:19: Error: The method 'DiscoveryQuery' isn't defined for the class '_MyHomePageState'.
 - '_MyHomePageState' is from 'package:sean_flutter_app/main.dart' ('lib/main.dart').
Try correcting the name to the name of an existing method, or defining a method named 'DiscoveryQuery'.
```

njooma: You can import it from `viam_sdk/protos/...` or you can add it as a `typedef` here similar to how we do it for `CloudMetadata` a few lines above (the preferred way IMO)

hexbabe: Got it let me check out the latter way

hexbabe: I tried the latter way, I had a hard time figuring out how to initialize the `extra` pb struct. Ended up importing `import 'package:viam_sdk/src/utils.dart';` for the map to struct helper logic. This seems like bad practice though since `utils.dart` seems purposefully excluded as an export from `viam_sdk.dart`. Open to ideas to making it easier to use `extra` without the `Struct get extraStruct => extra.toStruct();` utility in the `DiscoveryQuery` first draft.

randhid (approved):

njooma (approved):
Ok given the challenges around protobufs/struct, this is good as is

PR #286, opened 2024-10-30
RSDK-9180: Change TabularDataBySQL/MQL return type to raw BSON in Flutter SDK

jckras:
For context: The return type of the TabularDataBySQL/MQL proto has changed to return both a list of structs, the existing return type, and also a list of bytearrays that represent BSON.

Added a required dependency to bson to use the BSON library for dart
BSON provides a` BsonCodec.deserialize()` and `BsonCodec.serialize()` function for converting lists of byte sequences into lists of dart objects.

Testing:
Updated the `data` testing input to include `DateTime` objects, as the Flutter SDK will receive date fields in this format. I added a check to ensure that `DateTime` inputs are returned as native Flutter `DateTime` objects

I ensured that date fields are returned as native dart DateTime objects instead of strings by running tabularDataByMQL against real tabular data on viam-dev and printed the decoded data and their types. As you can see below the date and time elements are of type "DateTime":
Screenshot 2024-10-31 at 4 20 47 PM

### DIFF
--- a/lib/src/app/data.dart
+++ b/lib/src/app/data.dart
@@ -89,7 +89,6 @@ class DataClient {
       ..organizationId = organizationId
       ..sqlQuery = query;
     final response = await _dataClient.tabularDataBySQL(request);
-    // return response.data.map((e) => e.toMap()).toList();
     return response.rawData.map((e) => BsonCodec.deserialize(BsonBinary.from(e))).toList();
   }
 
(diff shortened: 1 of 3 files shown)


purplenicole730 (approved):

purplenicole730:
Is this considered a breaking change?

jckras:
> Is this considered a breaking change?

Yes! we talked off-line about this but yes since the dates/times will be given back as native Date objects and not strings this is a breaking change. I posted in eng-breaking-changes

stuqdog (approved):

PR #294, opened 2024-11-11
Convert sensor reading values to types

njooma:
Similar to how we do it in python: https://github.com/viamrobotics/viam-python-sdk/blob/8a1aef023ef2d77c4988c60e9de678aaac8e5bb8/src/viam/utils.py#L202

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -103,7 +103,6 @@ extension GetReadingsResponseUtils on GetReadingsResponse {
           case 'euler':
             primValue = Orientation_EulerAngles(roll: value['roll'], pitch: value['pitch'], yaw: value['yaw']);
           case 'vector3':
-            primValue = Vector3(x: value['x'], y: value['y'], z: value['z']);
           case 'angular_velocity':
             primValue = Vector3(x: value['x'], y: value['y'], z: value['z']);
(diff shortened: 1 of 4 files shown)


lib/src/utils.dart
```diff
@@ -92,6 +94,28 @@ grpc_duration.Duration durationToProto(Duration duration) {
...
+    return readings.map((key, value) => MapEntry(key, value.toPrimitive())).map((key, value) {
+      if (value is Map<String, dynamic> && value.keys.contains('_type')) {
+        dynamic primValue;
+        switch (value['_type']) {
+          case 'euler':
+            primValue = Orientation_EulerAngles(roll: value['roll'], pitch: value['pitch'], yaw: value['yaw']);
+          case 'vector3':
+            primValue = Vector3(x: value['x'], y: value['y'], z: value['z']);
+          case 'angular_velocity':
+            primValue = Vector3(x: value['x'], y: value['y'], z: value['z']);
```
stuqdog, line 108:
```suggestion
          case 'vector3':
          case 'angular_velocity':
            primValue = Vector3(x: value['x'], y: value['y'], z: value['z']);
```

stuqdog (approved):

PR #302, opened 2024-11-21
[hotfix] resource names

njooma:
This updates the way we create resources names when only given a name. It also updates the resource manager to look for short names when long names don't return anything


lib/src/resource/manager.dart
```diff
@@ -21,7 +21,18 @@ class ResourceManager {
 
   /// Get a resource with the given [ResourceName]
   T getResource<T>(ResourceName name) {
-    final resource = resources[name];
+    Resource? resource;
+    if (resources.containsKey(name)) {
+      resource = resources[name];
+    } else {
+      final resourcesWithoutRemotes = resources.map((rn, res) {
```
stuqdog, line 28: (q) Potentially out of scope for a hotfix but, should `resourcesWithoutRemotes` live on the `ResourceManager` such that we don't have to construct this new map potentially every time we call `getResource`?

stuqdog (approved):

lib/src/resource/manager.dart
njooma: Did it anyway -- it's a good thought

stuqdog (approved):

PR #308, opened 2024-11-27
[RSDK-9250] - Add stream options handlers

seanavery:
## Description

Adds handlers to the stream and robot client to hit options endpoints.

## Tests

Tested using local flutter web app that fires the following code block.

```dart
    final machine = await RobotClient.atAddress(
      host,
      RobotClientOptions.withApiKey(apiKeyID, apiKey),
    );
    final res = await machine.getStreamOptions("wc-cam");
    print(res);
    await machine.setStreamOptions("wc-cam", res[1].width, res[1].height);
    await machine.resetStreamOptions("wc-cam");
    await machine.close();
```

### DIFF
--- a/lib/src/media/stream/client.dart
+++ b/lib/src/media/stream/client.dart
@@ -85,6 +85,43 @@ class StreamManager {
     return client;
   }
 
+  Future<List<Resolution>> getStreamOptions(String name) async {
+    final sanitizedName = _getValidSDPTrackName(name);
+    try {
+      final response = await _client.getStreamOptions(GetStreamOptionsRequest()..name = sanitizedName);
+      _logger.d('Got options for stream named $name');
+      return response.resolutions;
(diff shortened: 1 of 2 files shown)


stuqdog (approved):

kevin49999:
@seanavery is there a way you could see what options a stream is currently running with?

seanavery:
@kevin49999 Currently there is not an easy way to look this up with an API call.. Would this be helpful for the frontend?

kevin49999:
> @kevin49999 Currently there is not an easy way to look this up with an API call.. Would this be helpful for the frontend?

It could be helpful, right now we will have a few redundant `setStreamOptions` calls. Though it doesn't seem like it's a big deal.

If we could read the current options I was thinking of using that to avoid calling set stream if it's already been done

PR #312, opened 2024-12-11
Start mdns after setting listeners

njooma:
According to the [docs](https://pub.dev/packages/bonsoir), you need to start mDNS AFTER setting listeners


stuqdog (approved):

acmorrow (approved):
I don't know enough about dart to take a position on the details of the code, but this does seem to bring it into alignment with the documented way of using the API. I also tested this locally, and with this change in place, a flutter app is able to discover a micro-rdk instance via mDNS running in the same isolated network, and then connect to it with WebRTC. Without this change, it doesn't ever manage to discover it.

LGTM and thanks for the quick fix!

PR #330, opened 2025-01-22
Add includeBinary to data funcs

njooma:
**BREAKING CHANGE**
We were missing the param to actually download the data. In the process of changing that, I had to change the return type of `binaryDataByIDs` since the user can request just the count now. 

As a flyby, I fixed `binaryDataByFilter` to actually use `countOnly`

### DIFF
--- a/test/unit_test/app/data_client_test.dart
+++ b/test/unit_test/app/data_client_test.dart
@@ -1,5 +1,3 @@
-import 'dart:typed_data';
-
 import 'package:bson/bson.dart' hide Timestamp;
 import 'package:fixnum/fixnum.dart';
 import 'package:flutter_test/flutter_test.dart';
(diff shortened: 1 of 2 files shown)


stuqdog (approved):

purplenicole730 (approved):

test/unit_test/app/data_client_test.dart
```diff
@@ -135,11 +139,16 @@ void main() {
...
-        when(serviceClient.binaryDataByIDs(any)).thenAnswer((_) => MockResponseFuture.value(BinaryDataByIDsResponse()..data.addAll(data)));
+        when(serviceClient.binaryDataByIDs(any)).thenAnswer((invocation) {
+          includeBinary = (invocation.positionalArguments[0] as BinaryDataByIDsRequest).includeBinary;
+          return MockResponseFuture.value(BinaryDataByIDsResponse()..data.addAll(data));
+        });
 
-        final response = await dataClient.binaryDataByIds(ids);
-        expect(response, equals(data));
+        final response = await dataClient.binaryDataByIds(ids, includeBinary: true);
+        expect(response.data, equals(data));
```
purplenicole730, line 150: Should we not test `response.count` as well now that the response has changed?

PR #341, opened 2025-02-18
RSDK-9898 - fix switch rename

stuqdog:


njooma (approved):
The `sed` separator argument doesn't have to be `/`! Improves readability, I like `#`, @lia-viam likes `@`: `s#foo#bar#g`

stuqdog:
> The `sed` separator argument doesn't have to be `/`! Improves readability, I like `#`, @lia-viam likes `@`: `s#foo#bar#g`

🤯 til!

PR #348, opened 2025-03-06
feat: add Switch component, ViamSwitchWidget

HipsterBrown:
While working on a recent project that uses a switch component from a module, I wanted to control the component from the Viam app. This PR adds the Switch implementation to the SDK as well as a ViamSwitchWidget to make it 🤞 quick to add to the mobile app. I tested out the component and widget by adding a new screen to the robot_example_app. 

Preview from robot_example_app running on MacOS:
Screenshot 2025-03-06 at 4 49 50 PM

I opted for using the SliderTheme in the ViamSwitchWidget instead of a row of buttons like the web app currently uses because it felt more natural given the variable amount of positions available to the component, meaning it will scale more easily to a mobile UI.

Open Question:
I noticed the updated import path for the generated proto code from "switch" to "nswitch" to avoid conflicts with the `switch` statement in Dart. I chose to keep using `Switch` as the name of the component and use `nswitch` for any class properties and worked around issues with Material by ignoring its included Switch widget class in the relevant files. Not sure if this is the best solution so I'm open to feedback to make sure the SDK usage is consistent with the other languages while not feeling foreign to Flutter devs.


example/viam_robot_example_app/lib/screens/switch.dart
```diff
@@ -0,0 +1,118 @@
...
+//                           child: Text('$index'),
+//                         );
+//                       })),
+//             ],
+//           ),
+//         ),
+//       ),
+//     );
+//   }
+// }
```
clintpurser, line 118: remove?

lib/src/components/switch/switch.dart
```diff
@@ -0,0 +1,61 @@
+import '../../gen/common/v1/common.pb.dart';
+import '../../resource/base.dart';
+import '../../robot/client.dart';
+
+/// {@category Components}
+/// Switch represents a set of finite positions.
+///
+/// For more information, see [Switch component](https://docs.viam.com/dev/reference/apis/components/switch/).
+abstract class Switch extends Resource {
```
clintpurser, line 9: yeah i see what you mean, `switch` is a overloaded term. I think it's fine to leave as is and let users handle the overloaded class name via imports, because we don't name anything else prefixed with `viam` or like `nSensor` - @njooma any thoughts on this?

clintpurser (approved):

lib/src/components/switch/switch.dart
njooma: I think it's fine to have it as `Switch`. We had to make it `nswitch` because proto was complaining, but I just tested it again with just `switch` and it seems to work, so I think maybe a proto update fixed it? I can update the protos in a different PR though

njooma (approved):

PR #351, opened 2025-03-07
RSDK-9900 - always find unused port for testing

stuqdog:
Changes the logic for finding a port for tests to ensure that the found port is unused.

It's hard to be absolutely confident this is right, since the failure is flaky and hard to reproduce reliably. But, the logic of the change seems sound to me, and is cribbed from logic [used by dart internally](https://chromium.googlesource.com/external/github.com/dart-lang/test/+/master/pkgs/test_core/lib/src/util/io.dart#147) for testing. I figure if it's good enough for the dart team, it's probably good enough for us!

### DIFF
--- a/test/test_utils.dart
+++ b/test/test_utils.dart
@@ -3,7 +3,7 @@ import 'dart:io';
 import 'package:flutter/material.dart';
 
 Future<int> getUnusedPort() {
-    return ServerSocket.bind(InternetAddress.anyIPv4, 0).then((socket) {
+  return ServerSocket.bind(InternetAddress.loopbackIPv4, 0).then((socket) {
     final port = socket.port;
     socket.close();
     return port;
--- a/test/test_utils.dart
+++ b/test/test_utils.dart
@@ -2,21 +2,22 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:flutter/material.dart';
+import 'package:grpc/grpc.dart';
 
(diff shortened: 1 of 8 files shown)


test/test_utils.dart
```diff
@@ -1,6 +1,31 @@
...
 
-int generateTestingPortFromName(String name) => 50000 + (name.hashCode % 10000);
+Future<int> getUnusedPort() async {
+  int? value;
+  await Future.doWhile(() async {
+    value = await ServerSocket.bind(InternetAddress.loopbackIPv4, await getUnsafeUnusedPort()).then((socket) {
+      final port = socket.port;
+      socket.close();
+      return port;
+    });
```
njooma, line 13: Do we need to test the bind here, since the socket binding is already being tested in `getUnsafeUnusedPort`?

stuqdog: Yeah, good question. I think I was misunderstanding the code I cribbed from, it adds this additional step but returns a `T` instead, the goal is to avoid a race condition that occurs when some other process binds to the port returned by `getUnsafeUnusedPort` after it returns. So basically repeatedly grab a _probably_ unused port then attempt to actually do something with it and then when you are able to do so successfully, return it. 

But since the issue we're running into is an error being thrown rather than a null return, this won't actually work smoothly without a bit more work. Will revisit and re-request review.

njooma (approved):

PR #354, opened 2025-03-13
RSDK-9901 - use git-access-token

stuqdog:
analogized from python PR

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -14,6 +14,8 @@ jobs:
       image: ghcr.io/cirruslabs/flutter:latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GIT_ACCESS_TOKEN }}
       - uses: bufbuild/buf-setup-action@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_protos.yml
```diff
@@ -14,6 +14,8 @@ jobs:
       image: ghcr.io/cirruslabs/flutter:latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
njooma, line 18: Other PRs use repo read token?

stuqdog: Yeah, flutter doesn't have access to the `REPO_READ_TOKEN`, but 1) they should have approximately the same permissions, and 2) I'm not actually sure this part is even necessary. It's a bit surprising to me that this shows up as a diff at all here because this part was already merged in when this branch was merged the other day. But either way, the most important change is changing the token of the Open PR step, as it changes _who_ opens the PR from `git-actions` to `viambot` and therefore changes the capacity to run tests.

njooma (approved):

PR #357, opened 2025-03-17
finish discovery removal, change alpha setting for color

stuqdog:
Finish removal of `Discovery` types to allow for compilation to finish.

Also changes call for setting of a color's `alpha` value in the switch widget to be a setting of `opacity` instead (in line with how it's done in the `motor` widget).

### DIFF
--- a/lib/widgets/resources/switch.dart
+++ b/lib/widgets/resources/switch.dart
@@ -66,8 +66,8 @@ class _ViamSwitchWidgetState extends State<ViamSwitchWidget> {
     return SliderThemeData(
       trackHeight: 10,
       thumbShape: const RoundSliderThumbShape(enabledThumbRadius: 10, disabledThumbRadius: 10),
-      activeTrackColor: Theme.of(context).colorScheme.secondary.withAlpha(64),
-      inactiveTrackColor: Theme.of(context).colorScheme.secondary.withAlpha(64),
+      activeTrackColor: Theme.of(context).colorScheme.secondary.withOpacity(0.25),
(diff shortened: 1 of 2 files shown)


lib/widgets/resources/switch.dart
```diff
@@ -66,8 +66,8 @@ class _ViamSwitchWidgetState extends State<ViamSwitchWidget> {
     return SliderThemeData(
       trackHeight: 10,
       thumbShape: const RoundSliderThumbShape(enabledThumbRadius: 10, disabledThumbRadius: 10),
-      activeTrackColor: Theme.of(context).colorScheme.secondary.withValues(alpha: 0.25),
-      inactiveTrackColor: Theme.of(context).colorScheme.secondary.withValues(alpha: 0.25),
+      activeTrackColor: Theme.of(context).colorScheme.secondary.withOpacity(0.25),
+      inactiveTrackColor: Theme.of(context).colorScheme.secondary.withOpacity(0.25),
```
stuqdog, line 70: cc @HipsterBrown, `withValues` doesn't seem to exist on a `Color` currently. I changed this to `withOpacity` to be in line with what `motor` does but I don't actually have any real feelings or thoughts on this, we could certainly use `withAlpha` instead if you want to keep setting the `alpha` value (though that function asks for a u8 int, not a float).

HipsterBrown: That's odd since `withOpacity` is deprecated in Flutter 3.27.0: https://docs.flutter.dev/release/breaking-changes/wide-gamut-framework#opacity

I don't know what our minimum supported version is for the Flutter SDK, so maybe we need to stick with `withOpacity` for a while.

PR #358, opened 2025-03-17
[CONSULT-940] add listMachineSummaries to flutter sdk

jckras:
(https://viam.atlassian.net/browse/CONSULT-940) - This PR adds a flutter sdk wrapper for listMachineSummaries. All of the changes in test/unit_test/mocks/service_clients_mocks.mocks.dart are generated code changes in order to build out a test in app_client_test.dart. In addition to adding tests to `app_client_test.dart` I tested locally using the viam_robot_example_app. Here is a video below:

https://github.com/user-attachments/assets/6b832a6d-8782-415c-8642-9692eec34225

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -18,6 +18,15 @@ class AppClient {
 
   AppClient(this._client);
 
+  /// List machine summaries for an organization
+  /// 
+  /// For more information, see [Fleet Management API] ](https://docs.viam.com/appendix/apis/fleet/).
+  Future<Iterable<LocationSummary>> listMachineSummaries(String organizationId) async {
+    final request = ListMachineSummariesRequest()..organizationId = organizationId;
+    final ListMachineSummariesResponse response = await _client.listMachineSummaries(request);
+    return response.locationSummaries;
+  }
+
   /// Get the id of the user with the email provided
   ///
   /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
--- a/test/unit_test/app/app_client_test.dart
+++ b/test/unit_test/app/app_client_test.dart
@@ -23,6 +23,37 @@ void main() {
   });
 
   group('App RPC Client Tests', () {
+    test('listMachineSummaries', () async {
+       final expected = [
+        LocationSummary()
+          ..locationId = 'id'
+          ..locationName = 'name'
+          ..machineSummaries.addAll([
+            MachineSummary()
+              ..machineId = 'machineId'
+              ..machineName = 'machineName'
+              ..partSummaries.addAll([
+                PartSummary()
+                  ..partId = 'partId'
+                  ..partName = 'partName'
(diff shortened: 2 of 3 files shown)


lib/src/app/app.dart
```diff
@@ -18,6 +18,15 @@ class AppClient {
 
   AppClient(this._client);
 
+  /// List machine summaries for an organization
+  /// 
+  /// For more information, see [Fleet Management API] ](https://docs.viam.com/appendix/apis/fleet/).
+  Future<Iterable<LocationSummary>> listMachineSummaries(String organizationId) async {
```
stuqdog, line 24: (q) Out of curiosity, why an `Iterable`? It looks like `Future<List<...>>` is the more common pattern when returning a list of data from a response type.

lib/src/app/app.dart
```diff
@@ -18,6 +18,15 @@ class AppClient {
...
+    final request = ListMachineSummariesRequest()..organizationId = organizationId;
+    final ListMachineSummariesResponse response = await _client.listMachineSummaries(request);
+    return response.locationSummaries;
```
stuqdog, line 27: This is a little strange to me. The name of the method is `listMachineSummaries` and the proto matches that, but what we're actually returning are _location_ summaries, not machine summaries. It's unfortunate that we have to choose between consistency with proto and precision in how what we communicate to users, but personally I think we should err on the side of improving the user experience and rename this to `listLocationSummaries`.

Admittedly the location summaries contain machine summaries in them so it's not totally dishonest; I'm definitely willing to be convinced otherwise here.

stuqdog (commented):
A couple high level thoughts but generally this looks good to me!

lib/src/app/app.dart
jckras: @clintpurser and I went back and forth on this. We just figured that since the proto returned an iterable of LocationSummaries that we would keep it as an iterable. But not opposed to changing this to a list for consistency. good call

lib/src/app/app.dart
jckras: this is a really good point. I don't have a great reason for keeping the name as `listMachineSummaries`  other than that it matches the proto - which I like. If we keep it the same name as the proto I feel that reduces confusion when referencing the proto or potentially other place we might use this api call.

stuqdog: Hmm... yeah I think at a minimum we should have a clarifying comment. I am still a little worried that maintaining proto consistency makes things easer and less confusing for us at the cost of making things _more_ confusing for a user. It's kind of a weird pattern that a user would read the function signature and get one impression of what this returns, but then going to the comment tells them that actually it's returning something else.

That said, it also makes sense that machines would be grouped by location, and we _do_ already have a `listLocations` API for accessing location data, so it makes sense that this would really just be used primarily for machine info. So I think it's probably okay to leave as-is, so long as we clarify how the `locationSummaries` actually get us to `machineSummaries`.

PR #359, opened 2025-03-18
[CONSULT-940] add listMachineSummaries to flutter sdk

jckras:
(https://viam.atlassian.net/browse/CONSULT-940) - This PR adds a flutter sdk wrapper for listMachineSummaries. All of the changes in `test/unit_test/mocks/service_clients_mocks.mocks.dart` are generated code changes in order to build out a test in app_client_test.dart. In addition to adding tests to `app_client_test.dart` I tested locally using the viam_robot_example_app. 

@stuqdog, i updated the comment to be more clear about the what we are actually returning, but let me know if you still prefer to change name of function, not opposed

Here is a video below:

https://github.com/user-attachments/assets/6b832a6d-8782-415c-8642-9692eec34225

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -18,7 +18,7 @@ class AppClient {
 
   AppClient(this._client);
 
-  /// List LocationSummary objects which contain machine summaries within them for an organization
+  /// List all [MachineSummary]s for an organization grouped by Location
   ///
   /// For more information, see [Fleet Management API] ](https://docs.viam.com/appendix/apis/fleet/).
   Future<List<LocationSummary>> listMachineSummaries(String organizationId) async {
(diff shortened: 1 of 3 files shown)


lib/src/app/app.dart
```diff
@@ -18,6 +18,15 @@ class AppClient {
 
   AppClient(this._client);
 
+  /// List LocationSummary objects which contain machine summaries within them for an organization
```
stuqdog, line 21: Hmm... sorry for the back and forth on this! I think the important part of this is the machine summaries (`listLocations` exists if someone wants location data specifically), but we want to address the fact that they're returned as `LocationSummary`s. Maybe something like `/// List all [MachineSummary]s, grouped by location.`?

martha-johnston (approved):
lgtm from a mobile perspective! good work Julie

stuqdog (approved):
nice, lgtm!

PR #360, opened 2025-03-18
[RSDK-10127] Add get and update metadata functions

JosephBorodach:
# Description
Add user-defined metadata CRUD operations 

**See ticket [here](https://viam.atlassian.net/jira/software/c/projects/APP/boards/64?selectedIssue=RSDK-10127)**

### DIFF
--- a/lib/src/app/app.dart
+++ b/lib/src/app/app.dart
@@ -744,7 +744,7 @@ class AppClient {
   Future<void> updateOrganizationMetadata(String id, Map<String, dynamic> data) async {
     final request = UpdateOrganizationMetadataRequest()
       ..organizationId = id
-      ..data = _mapToStruct(data);
+      ..data = data.toStruct();
     await _client.updateOrganizationMetadata(request);
   }
 
@@ -762,7 +762,7 @@ class AppClient {
   Future<void> updateLocationMetadata(String id, Map<String, dynamic> data) async {
     final request = UpdateLocationMetadataRequest()
       ..locationId = id
-      ..data = _mapToStruct(data);
+      ..data = data.toStruct();
     await _client.updateLocationMetadata(request);
   }
 
@@ -780,7 +780,7 @@ class AppClient {
   Future<void> updateRobotMetadata(String id, Map<String, dynamic> data) async {
     final request = UpdateRobotMetadataRequest()
       ..id = id
-      ..data = _mapToStruct(data);
+      ..data = data.toStruct();
     await _client.updateRobotMetadata(request);
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/app/app.dart
```diff
@@ -729,4 +729,76 @@ class AppClient {
     final request = CreateKeyFromExistingKeyAuthorizationsRequest()..id = id;
     return await _client.createKeyFromExistingKeyAuthorizations(request);
   }
+
+  /// Retrieves user-defined [Metadata] for an organization.
+  ///
+  /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
+  Future<GetOrganizationMetadataResponse> getOrganizationMetadata(String id) async {
```
stuqdog, line 736: (minor/nit) can we rename to `organizationId` or `orgId`? it's more specific and more consistent with similar calls elsewhere in the SDK.

Same comment applies to below methods.

lib/src/app/app.dart
```diff
@@ -729,4 +729,76 @@ class AppClient {
...
+    return await _client.getOrganizationMetadata(request);
+  }
+
+  /// Updates user-defined [Metadata] for an organization.
+  ///
+  /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
+  Future<void> updateOrganizationMetadata(String id, Map<String, dynamic> data) async {
+    final request = UpdateOrganizationMetadataRequest()
+      ..organizationId = id
+      ..data = _mapToStruct(data);
```
stuqdog, line 747: `_mapToStruct` is undefined, which is causing errors. I think what you're looking for is `data.toStruct()`.

test/unit_test/app/app_client_test.dart
```diff
@@ -754,5 +754,12 @@ void main() {
       final response = await appClient.createKeyFromExistingKeyAuthorizations('id');
       expect(response, equals(expected));
     });
+
+    test('getOrganizationMetadata', () async {
+      final expected = GetOrganizationMetadataResponse()..data = (Struct()..fields['key'] = (Value()..stringValue = 'value'));
+      when(serviceClient.getOrganizationMetadata(any)).thenAnswer((_) => MockResponseFuture.value(expected));
+      final response = await appClient.getOrganizationMetadata('orgId');
+      expect(response, equals(expected));
+    });
```
stuqdog, line 763: This test looks good, but we should probably add tests for all the other methods you implemented as well!

stuqdog (changes requested):

lib/src/app/app.dart
JosephBorodach: great catch, thanks for calling this out

stuqdog (approved):
lgtm!

njooma (approved):

PR #379, opened 2025-04-24
add last known update to updateRobotPart and updateFragment

gmulz:


lib/src/app/app.dart
```diff
@@ -363,11 +363,12 @@ class AppClient {
   /// Update a specific [RobotPart] by ID
   ///
   /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
-  Future<RobotPart> updateRobotPart(String partId, String name, Map<String, dynamic> robotConfig) async {
+  Future<RobotPart> updateRobotPart(String partId, String name, Map<String, dynamic> robotConfig, {DateTime? lastKnownUpdate}) async {
     final updateRobotPartRequest = UpdateRobotPartRequest()
       ..id = partId
       ..name = name
       ..robotConfig = robotConfig.toStruct();
+    if (lastKnownUpdate != null) updateRobotPartRequest.lastKnownUpdate = lastKnownUpdate;
```
stuqdog, line 371: I think this is causing tests to fail because the request is expecting a `Timestamp` but getting a `DateTime`.

Same in `updateFragment` below.

gmulz: thanks! yeah i saw the error but got busy with some other stuff, just looping back on the fix now

stuqdog (approved):
looks like tests are still failing :( once those are fixed, lgtm!

gmulz:
> looks like tests are still failing :( once those are fixed, lgtm!

yeah i forgot an import, they are passing now. i don't know how to use dart lol

PR #385, opened 2025-05-14
Hardcode ip for provisioning, fix client used with apikey

kevin49999:
Updates to use the ip `10.42.0.1` instead of a hostname `viam.setup` during provisioning. Using `viam.setup` while using cellular (but connected to the hotspot) on iOS does not work and the requests fail.

Also,

Testing in my [example project](https://github.com/viamrobotics/viam_flutter_provisioning_widget/tree/main/example/hotspot_provisioning), I couldn't get the init w/ api key to function (for provisioning): https://github.com/viamrobotics/viam_flutter_provisioning_widget/pull/2

I think the host `provisioning.viam` in this init for the `ProvisioningClient` is outdated / doesn't match how it's setup when using an access token (uses `viam.setup`). Plus we want to use an ip now here as well

### DIFF
--- a/lib/src/app/provisioning.dart
+++ b/lib/src/app/provisioning.dart
@@ -7,8 +7,6 @@ enum NetworkType { wifi, wired }
 
 /// {@category Viam SDK}
 /// gRPC client for connecting to Viam's Provisioning Service
-///
-/// All calls must be authenticated.
 class ProvisioningClient {
   final ProvisioningServiceClient _client;
 
--- a/lib/src/viam_sdk_impl.dart
+++ b/lib/src/viam_sdk_impl.dart
@@ -31,7 +31,14 @@ class ViamImpl implements Viam {
     _billingClient = BillingClient(BillingServiceClient(_clientChannelBase));
     _dataClient = DataClient(
         DataServiceClient(_clientChannelBase), DataSyncServiceClient(_clientChannelBase), DatasetServiceClient(_clientChannelBase));
-    _provisioningClient = ProvisioningClient(ProvisioningServiceClient(ClientChannel('provisioning.viam', port: 4772)));
+
+    _provisioningClient = ProvisioningClient(ProvisioningServiceClient(
+      ClientChannel(
+        'viam.setup',
+        port: 4772,
+        options: const ChannelOptions(credentials: ChannelCredentials.insecure()),
+      ),
+    ));
   }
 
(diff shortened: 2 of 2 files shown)


lib/src/app/provisioning.dart
```diff
@@ -7,8 +7,6 @@ enum NetworkType { wifi, wired }
 
 /// {@category Viam SDK}
 /// gRPC client for connecting to Viam's Provisioning Service
-///
-/// All calls must be authenticated.
```
kevin49999, line 11: I think this is the one client where this doesn't apply

lib/src/viam_sdk_impl.dart
```diff
@@ -44,7 +51,7 @@ class ViamImpl implements Viam {
 
     _provisioningClient = ProvisioningClient(ProvisioningServiceClient(
       ClientChannel(
-        'viam.setup',
+        '10.42.0.1',
```
kevin49999, line 54: @ale7714 had the idea to change this and the context for why it's necessary:
Screenshot 2025-05-15 at 10 46 27 AM

clintpurser: we've made these changes with not breaking android in mind as well, right?

clintpurser (approved):

lib/src/viam_sdk_impl.dart
```diff
@@ -31,7 +31,14 @@ class ViamImpl implements Viam {
     _billingClient = BillingClient(BillingServiceClient(_clientChannelBase));
     _dataClient = DataClient(
         DataServiceClient(_clientChannelBase), DataSyncServiceClient(_clientChannelBase), DatasetServiceClient(_clientChannelBase));
-    _provisioningClient = ProvisioningClient(ProvisioningServiceClient(ClientChannel('provisioning.viam', port: 4772)));
```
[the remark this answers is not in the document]
kevin49999: When using an access token to authenticate, it was always `viam.setup`

I only noticed this (and it didn't work) when trying to init w/ an api key (uses a different init). I think this has been wrong but no one uses this (or our apps don't use it)

lib/src/viam_sdk_impl.dart
ale7714: BTW this should not be ios specific. This is about how network stacks in computers handle multiple interfaces. I can see how my comment makes this sounds OS specific tho so sorry for the confusion.

njooma (approved):

PR #411, opened 2025-08-01
Add support for retrieving and displaying Switch labels

Joyceku1020:
- Added getLabels() to Switch interface and SwitchClient

- Updated ViamSwitchWidget to display labels if available

### DIFF
--- a/lib/src/components/switch/switch.dart
+++ b/lib/src/components/switch/switch.dart
@@ -30,11 +30,20 @@ abstract class Switch extends Resource {
   /// Get the number of available positions (int) of the [Switch].
   ///
   /// ```
-  /// await myServo.getNumberOfPositions();
+  /// await mySwitch.getNumberOfPositions();
   /// ```
   ///
   /// For more information, see [Switch component](https://docs.viam.com/dev/reference/apis/components/servo/#getnumberofpositions).
   Future<int> getNumberOfPositions({Map<String, dynamic>? extra});
+
+  /// Get the list of labels associated with each position of the [Switch].
+  ///
+  /// ```
+  /// var labels = await mySwitch.getLabels();
+  /// ```
+  ///
+  /// For more information, see [Switch component](https://docs.viam.com/dev/reference/apis/components/servo/#getnumberofpositions).
+  Future<List<String>> getLabels({Map<String, dynamic>? extra});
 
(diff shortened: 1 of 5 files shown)


lib/src/components/switch/switch.dart
```diff
@@ -30,12 +30,21 @@ abstract class Switch extends Resource {
   /// Get the number of available positions (int) of the [Switch].
   ///
   /// ```
-  /// await myServo.getNumberOfPositions();
+  /// await mySwitch.getNumberOfPositions();
   /// ```
   ///
   /// For more information, see [Switch component](https://docs.viam.com/dev/reference/apis/components/servo/#getnumberofpositions).
   Future<int> getNumberOfPositions({Map<String, dynamic>? extra});
```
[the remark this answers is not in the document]
Joyceku1020: Thanks for the suggestion! Since Dart 3 supports tuples, we can return both numberOfPositions and labels as a (int, List<String>) without needing a custom class.

test/unit_test/components/switch_test.dart
```diff
@@ -73,6 +79,11 @@ void main() {
       expect(testResult, nswitch.position);
     });
 
+    test('getLabels should return the expected labels', () async {
+      final labels = await nswitch.getLabels();
+      expect(labels, nswitch.labels);
+    });
+
```
jckras, line 86: great test

lib/src/components/switch/switch.dart
```diff
@@ -6,6 +6,13 @@ import '../../robot/client.dart';
 /// Switch represents a set of finite positions.
 ///
 /// For more information, see [Switch component](https://docs.viam.com/dev/reference/apis/components/switch/).
+class PositionsInfo {
+  final int numberOfPositions;
+  final List<String>? labels;
+
+  PositionsInfo({required this.numberOfPositions, this.labels});
+}
+
```
njooma, line 15: Because this class mirrors the attributes in `GetNumberOfPositionsResponse`, it's probably easier to typealias (see [camera example](https://github.com/viamrobotics/viam-flutter-sdk/blob/98ea2d6eb538d76eb137cdfbd677bb6cee135242/lib/src/components/camera/camera.dart#L9))

njooma (changes requested):

jckras:
thanks for completing this @njooma

PR #425, opened 2025-09-09
Allow users to dial with only `accessToken` and no `credentials`

clintpurser:
I was exploring using `accessToken` like the Svelte SDK does for dialing a robot and realized if we don’t include `credentials` and it is `Null` we go down a weird code path.

As an aside, I think GrpcOrGrpcWebClientChannel doesn’t work for a `StreamManager` which is what happens currently happens prior to this change if users only set `accessToken` and not `credentials`. This feels like a bug that should be investigated.


stuqdog (approved):
looks reasonable to me!

PR #519, opened 2026-03-31
Avoid multiple heartbeat tasks

kevin49999:
Based on #497 and pairing with @njooma (main change is his commit [#9473fc8](https://github.com/viamrobotics/viam-flutter-sdk/commit/9473fc8aaac76dd6aab1db8a025b60ebd3f77b20))

Tested with `viam_robot_example_app`


stuqdog (commented):
Generally looks reasonable to me though I think I'd defer to Naveed who likely has more context since you two paired on this!

Fwiw if change the runner for the test file to `ubuntu-latest` I think that should causes tests to run properly and would be a good flyby fix in general

njooma (approved):

PR #531, opened 2026-04-28
OAuth App user creation wrapper

gmulz:
wraps the new CreateOAuthAppUser rpc


njooma (approved):

lib/src/app/app.dart
```diff
@@ -833,6 +833,30 @@ class AppClient {
...
+  ///
+  /// For more information, see [Fleet Management API](https://docs.viam.com/appendix/apis/fleet/).
+  Future<CreateOAuthAppUserResponse> createOAuthAppUser({
+    required String orgId,
+    required String applicationId,
+    required String email,
+    required String firstName,
+    required String lastName,
+    required String password,
+  }) async {
```
kevin49999, line 849: May run into some issues here w/ calling this in third party apps - all these AppClient calls have to be authenticated first cc: @gmulz @njooma

PR #541, opened 2026-06-18
RSDK-11037 Add geometries to every component

ale7714:
## Summary

Part of the effort to expose `geometries` on every component that supports it (alongside `DoCommand`).

In the Flutter SDK, most components had an **unimplemented** `getGeometries` server handler (`throw UnimplementedError()`) and no client/interface method, so geometries could not be retrieved at all. This PR implements `getGeometries` end-to-end for every component whose API defines a `GetGeometries` RPC:

- audio input
- audio output
- base
- board
- camera
- generic
- gripper
- input controller
- motor
- movement sensor
- sensor
- servo

This brings them in line with **arm, encoder, and gantry**, which were already complete.

> Note: `button`, `power sensor`, and `switch` are intentionally excluded — their APIs do not define a `GetGeometries` RPC.

## Changes

For each component:
- **Interface** (`<name>.dart`): add `Future<List<Geometry>> getGeometries({Map<String, dynamic>? extra})`.
- **Client** (`client.dart`): implement the method by calling the `GetGeometries` RPC and returning `response.geometries`.
- **Service** (`service.dart`): implement the previously stubbed server handler, delegating to the resource.

The input controller client previously had a partial, non-`@override` `getGeometries` that returned the raw `GetGeometriesResponse`; it now returns `List<Geometry>` and is declared on the interface, consistent with every other component.

Unit tests were added for the fake, service, and client of each updated component (mirroring the existing arm/encoder/gantry geometries tests).

## Testing

- `flutter test test/unit_test/components/ --concurrency=1` → all 436 tests pass (the suite uses fixed ports and is flaky under the default parallel runner — unrelated to this change).
- `dart format --line-length=140 --set-exit-if-changed` → clean.
- `dart analyze ./lib` → no new issues.

Key: RSDK-11037

Co-authored by Claude agent for Jira.

### DIFF
--- a/lib/src/components/audio_in/audio_in.dart
+++ b/lib/src/components/audio_in/audio_in.dart
@@ -22,6 +22,13 @@ abstract class AudioIn extends Resource {
   /// Get the audio properties of this audio in device
   Future<GetPropertiesResponse> getProperties({Map<String, dynamic>? extra});
 
+  /// Get all geometries associated with the [AudioIn]
+  ///
+  /// ```
+  /// final geometries = await myAudioIn.getGeometries();
+  /// ```
+  Future<List<Geometry>> getGeometries({Map<String, dynamic>? extra});
+
   /// Get the [ResourceName] for this [AudioIn] with the given [name]
   static ResourceName getResourceName(String name) {
     return AudioIn.subtype.getResourceName(name);
--- a/lib/src/components/audio_in/client.dart
+++ b/lib/src/components/audio_in/client.dart
@@ -23,6 +23,15 @@ class AudioInClient extends AudioIn implements ResourceRPCClient {
   AudioInClient(this.name, this.channel);
 
   @override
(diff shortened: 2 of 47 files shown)


lib/src/components/audio_in/audio_in.dart
```diff
@@ -22,6 +22,13 @@ abstract class AudioIn extends Resource {
   /// Get the audio properties of this audio in device
   Future<GetPropertiesResponse> getProperties({Map<String, dynamic>? extra});
 
+  /// Get all geometries associated with the [AudioIn]
```
stuqdog, line 25: (total aside) is it actually correct that an AudioIn component could have geometries? I realize from a "the RPC is exposed" perspective the answer is yes so definitely makes sense to include here, but it just surprises me a bit!

stuqdog (approved):

PR #548, opened 2026-07-29
Clean up for v1.0 

martha-johnston:
This PR has the following changes in order to get this repo into a v1.0 state:
- Move widgets into their own example package outside of lib
- Keep button, camera stream, and multi camera stream accessible in lib (all used in other app)
- Rename example apps to be more descriptive
- Add robot location (necessary) to .env.template
- Update example app READMEs
- Add missing components to example apps so all component types (and vision service) are supported


lib/src/rpc/dial.dart
```diff
@@ -170,6 +170,11 @@ Future<ClientChannelBase> dial(String address, DialOptions? options, String Func
 
   if (opts.attemptMdns) {
     final mdnsSW = Stopwatch()..start();
+    // Remember the caller's WebRTC options so they can be restored if the mDNS
```
njooma, line 173: What problem is this solving? We already create a copy of the `dialOpts` in the `try` block

martha-johnston: dart copy doesn't work this way, it's not actually a copy so the name is misleading, so it was changing the actual dialOpts, and the original address that we need later gets lost. I'll change it so we don't use the unnecessary dialOptsCopy anymore

martha-johnston: kind of a flyby fix when testing the example apps but I'll dig in a little more and see if this is necessary to run the apps

njooma: Do you think it was the way we were doing this `final dialOptsCopy = opts.._usingMdns = true;`? This feels wrong to me, like we're updating original `opts._usingMdns` and then assigning that to `dialOptsCopy`. I think that needs to be split into 2 lines, but then maybe should work?

martha-johnston: we were always directly updating the original `opts` because the copy just copies the reference, it doesn't make a deep copy, we'd have to write that. 

my understanding is that we're storing the two values we need to reset after attempting an mdns connection, and then resetting them because that's all we need, rather than writing a deep copy. this error came about from testing with a simulator, it wouldn't connect because the `webRtcOptions` and `signalingServerAddress` were overwritten in the try block, but I think with an actual device, the try block succeeds, so we don't need to restore the original versions

martha-johnston: can look at this in a follow up to take a closer look, but I did just test and this fix is needed with a physical device as well

njooma (approved):

